PR #2, opened 2022-09-27
Development Environment Setup

gmacgre:
Set up the necessary items for the development of the bot, including:
TypeScript
node modules
ESLint
Prettier
Prisma
SQLite
dotenv
changelog
and a basic README

### DIFF
--- a/.env
+++ b/.env
@@ -1,7 +1,0 @@
-# Environment variables declared in this file are automatically made available to Prisma.
-# See the documentation for more detail: https://pris.ly/d/prisma-schema#accessing-environment-variables-from-the-schema
-
-# Prisma supports the native connection string format for PostgreSQL, MySQL, SQLite, SQL Server, MongoDB and CockroachDB.
-# See the documentation for all the connection string options: https://pris.ly/d/connection-strings
-
-DATABASE_URL="postgresql://johndoe:randompassword@localhost:5432/mydb?schema=public"
--- a/.env
+++ b/.env
@@ -1,0 +1,7 @@
+# Environment variables declared in this file are automatically made available to Prisma.
+# See the documentation for more detail: https://pris.ly/d/prisma-schema#accessing-environment-variables-from-the-schema
+
+# Prisma supports the native connection string format for PostgreSQL, MySQL, SQLite, SQL Server, MongoDB and CockroachDB.
+# See the documentation for all the connection string options: https://pris.ly/d/connection-strings
+
+DATABASE_URL="postgresql://johndoe:randompassword@localhost:5432/mydb?schema=public"
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,2 +1,0 @@
-node_modules
-dist
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,2 @@
+node_modules
+dist
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,12 @@
+{
+    "root": true,
+    "parser": "@typescript-eslint/parser",
+    "plugins": [
+      "@typescript-eslint"
+    ],
+    "extends": [
+      "eslint:recommended",
+      "plugin:@typescript-eslint/eslint-recommended",
+      "plugin:@typescript-eslint/recommended"
+    ]
+  }
--- a/.eslintrc
+++ b/.eslintrc
@@ -47,5 +47,6 @@
       "space-unary-ops": "error",
       "spaced-comment": "error",
       "yoda": "error"
-    }
+    },
+    "ignorePatterns": ["dist/"]
   }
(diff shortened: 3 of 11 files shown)


.env
```diff
@@ -0,0 +1,7 @@
+# Environment variables declared in this file are automatically made available to Prisma.
```
AverageHelper, line 1: Please add .env to the .gitignore file

AverageHelper (changes requested):

.gitignore
```diff
@@ -0,0 +1,4 @@
+node_modules
+node_modules/*
+dist
+dist/*
```
SpencerHastings, line 4: I don't think you need to duplicate these.

JstnMcBrd: I'll fix that too.

package.json
```diff
@@ -0,0 +1,34 @@
...
+        "url": "git+https://github.com/BYU-CS-Discord/CSBot.git"
+    },
+    "author": "",
+    "license": "Unlicense",
+    "bugs": {
+        "url": "https://github.com/BYU-CS-Discord/CSBot/issues"
+    },
+    "homepage": "https://github.com/BYU-CS-Discord/CSBot#readme",
+    "devDependencies": {
+        "@types/sqlite3": "^3.1.8",
```
SpencerHastings, line 20: Not sure, but figured I'd check, are these sqlite types necessary?

.env
JstnMcBrd: Done.

CHANGELOG.md
```diff
@@ -0,0 +1 @@
+Set up an environment for the project.
```
SpencerHastings, line 1: Something like this format might be nice https://keepachangelog.com/en/1.0.0/

package.json
AverageHelper: They aren't for Prisma

SpencerHastings: Oh there's a sqlite3 dependency below, is that necessary?

JstnMcBrd: Should I remove that line?

SpencerHastings: If prisma takes care of the database connection I don't think sqlite is a necessary dependency, even if you use sqlite as the backend, idk for sure tho

JstnMcBrd: We'll get rid of it for now. Can always add it back later.

CHANGELOG.md
gmacgre: Done!

package.json
SpencerHastings: Yeah I'm pretty sure the sqlite dependency isn't necessary to use sqlite with prisma

.eslintignore
```diff
@@ -0,0 +1,2 @@
+node_modules
```
AverageHelper, line 1: Would this make more sense as an `ignorePatterns` array in `.eslintrc` instead? It's only the two lines, and both of these files deal with ESLint configuration.

AverageHelper (changes requested):

.eslintrc
```diff
@@ -0,0 +1,51 @@
...
+      "@typescript-eslint"
+    ],
+    "extends": [
+      "eslint:recommended",
+      "plugin:@typescript-eslint/eslint-recommended",
+      "plugin:@typescript-eslint/recommended"
+    ],
+    "rules": {
+      "arrow-spacing": ["warn", { "before": true, "after": true }],
+      "brace-style": ["error", "stroustrup", { "allowSingleLine": true }],
```
AverageHelper, line 14: We gotta bikeshed this. [Docs here.](https://eslint.org/docs/latest/rules/brace-style)

I prefer the default `1tbs` style. Should we take a poll on the others?

.eslintignore
gmacgre: Done. Also seems that it implicitly ignores node_modules
https://eslint.org/docs/latest/user-guide/configuring/ignoring-code

AverageHelper (changes requested):
I gotta read the rest of this ESLint config, you've changed a lot of defaults lol

package.json
```diff
@@ -0,0 +1,33 @@
...
+        "eslint": "^8.24.0",
+        "eslint-config-prettier": "^8.5.0",
+        "eslint-plugin-prettier": "^4.2.1",
+        "prettier": "^2.7.1",
+        "prisma": "^4.3.1",
+        "typescript": "^4.8.3"
+    },
+    "dependencies": {
+        "dotenv": "^16.0.2",
+        "sqlite3": "^5.1.1"
```
AverageHelper, line 31: Not needed with Prisma

AverageHelper (changes requested):
Ok, I think I pretty much agree with the rest of the ESLint config, apart from the block style noted in my earlier comment.

package.json
gmacgre: Done!

AverageHelper (approved):

Issue #8, opened 2022-09-27
Slash Command - basic Help


AverageHelper:
PR at #23

AverageHelper:
I think having this one in place would give us a good base on which to build the rest.

Plyb:
If we build this right, we shouldn't have to manually update the help text every time we add a new command. Can we structure our project in such a way that command descriptions are a necessary part of registering the command?

AverageHelper:
Absolutely! I'm working on a branch right now to demo just that

JstnMcBrd:
Every command file should have the name + description as long as we use the discord.js SlashCommandBuilder. It would be easy enough to load all the commands in `./commands/*.ts` and pull their names + descriptions.

PR #14, opened 2022-09-28
Basic command context

### LINKED ISSUE #7: Set up basic event and command handling
The suggested best practice for discord.js is pretty elegant.
[Command Handling](https://discordjs.guide/creating-your-bot/command-handling.html)
[Event Handling](https://discordjs.guide/creating-your-bot/event-handling.html)

AverageHelper:
### Added
- Discord client
- Event handling
- Command dispatcher
- Lots of unit tests
- Some fixes to the dev environment, making ESLint and Prettier fight less

The test coverage is "ok," but could be improved. I added a CI check to make coverage clear and keep us accountable.

Closes #7


AverageHelper:
I want at least a few eyes on this, since there's so much to go over. This code, to me, looks fine for now, but I wrote it. I'm bound to have missed something.

I'm happy with the overall structure as-is, but I'm happy to discuss before we merge.

AverageHelper:
I think #8 would be most easily solved after we get this dispatch work (or something like it) merged in

package.json
```diff
@@ -64,13 +68,13 @@
...
-		"jest": "29.0.3",
-		"jest-environment-node": "29.0.3",
+		"jest": "28.1.3",
+		"jest-environment-node": "28.1.3",
 		"jest-extended": "3.1.0",
 		"nodemon": "2.0.20",
 		"prettier": "2.7.1",
 		"prisma": "4.3.1",
-		"ts-jest": "29.0.2",
+		"ts-jest": "28.0.8",
```
JstnMcBrd, line 77: Just curious, why downgrade the versions of these packages? Do the newest versions have some kind of issue?

src/@types/CommandContext.d.ts
```diff
@@ -0,0 +1,103 @@
...
+	/** Information relevant to a command invocation. */
+	interface CommandContext {
+		/** The command invocation interaction. */
+		readonly interaction: CommandInteraction;
+
+		/** Our own signed-in Discord client. */
+		readonly client: Client<true>;
+
+		/** The place to submit informative debug messages. */
+		readonly logger: Console; // TODO: More extensible logger than `console`?
```
JstnMcBrd, line 25: We should start a thread about this on Discord.

src/@types/CommandContext.d.ts
```diff
@@ -0,0 +1,103 @@
...
+
+		/** Replies to the command invocation, optionally pinging the command's sender. */
+		reply: (
+			options:
+				| string
+				| Omit<MessageReplyOptions, 'flags'>
+				| (Omit<InteractionReplyOptions, 'flags'> & {
+						shouldMention?: boolean;
+				  })
+		) => Promise<void>;
```
JstnMcBrd, line 77: From my experience, discord.js doesn't allow you to specify whether to send a ping when replying to a message (like users can). Has that changed?

src/commands/help.ts
```diff
@@ -0,0 +1,8 @@
+export const help: GlobalCommand = {
+	name: 'help',
+	description: 'Prints the list of commands',
+	requiresGuild: false,
+	async execute({ reply }) {
+		await reply('Hello, world!');
```
JstnMcBrd, line 6:
```suggestion
		// TODO: finish this command
		await reply('Hello, world!');
```
Just for the todo tree :)

src/commands/index.ts
```diff
@@ -0,0 +1,17 @@
...
+	const name = cmd.name;
+
+	if (allCommands.has(name)) {
+		throw new TypeError(
+			`Failed to add command '${name}' when a command with that name was already added`
+		);
+	}
+
+	allCommands.set(name, cmd);
+}
```
JstnMcBrd, line 17: Couldn't this index just automatically compile the list of commands? I mean, all I has to do is use
`fs.readdirSync('./').filter(file => file.endsWith('.ts') && !file.endsWith('.test.ts' && file != 'index.ts');`
And then we don't have to update this every time we add a new command.

src/handleInteraction.ts
```diff
@@ -0,0 +1,192 @@
...
+ * @param logger The place to write system messages.
+ */
+export async function handleInteraction(
+	interaction: CommandInteraction,
+	logger: Console
+): Promise<void> {
+	// Don't respond to bots or ourselves
+	if (interaction.user.bot) return;
+	if (interaction.user.id === interaction.client.user.id) return;
+
```
JstnMcBrd, line 31: I'd recommend adding
```suggestion
if (!interaction.isChatInputCommand()) return;
```
because not every interaction is a command.

src/handleInteraction.ts
```diff
@@ -0,0 +1,192 @@
...
+					} else {
+						await interaction.followUp({ ...options, ephemeral: true });
+					}
+				} catch (error) {
+					logger.error('Failed to follow up on interaction:', error);
+				}
+			} else {
+				const reply = await replyPrivately(interaction, options, viaDM);
+				if (reply === false) {
+					logger.info(`User ${logUser(interaction.user)} has DMs turned off.`);
```
JstnMcBrd, line 107: In this situation, maybe send an ephemeral reply to the command telling the user "I can't access your DMs! Please allow messages from server members and try again."

JstnMcBrd: Nevermind, you included that elsewhere.

src/handleInteraction.ts
JstnMcBrd: Nevermind, you did this elsewhere.

JstnMcBrd (commented):
That was a lot to read but it looks good. We should probably get someone with more TypeScript + Jest experience to take a look at it too.

JstnMcBrd:
Did we want to abstract out the event handling system too? Right now, all the event handlers are hard-coded into [main.ts](https://github.com/BYU-CS-Discord/CSBot/blob/79a2e8bcd9e4eae0048170df5cc91e42082faec0/src/main.ts). Using a system similar to [discord.js's suggestion](https://discordjs.guide/creating-your-bot/event-handling.html) would be more flexible and consistent with the rest of the project.

package.json
AverageHelper: When I made the last PR, I didn't check the versions I installed, since I thought 28 was still under active development. 29 was fine for the one empty file, but threw strange warnings in the console about `ts-jest`. As I wrote more tests, I found that 29 disagreed with TypeScript a ton, and wouldn't consume the config the way I'd laid it down. Seems there were some breaking changes from 28 to 29.

The downgrade fixed all of the issues.

src/@types/CommandContext.d.ts
AverageHelper: That has! See the [`allowedMentions`](https://discordjs.guide/additional-info/changes-in-v13.html#allowed-mentions) object. That object has a property, `replied_user`, that determines whether a message should ping the user whose message is being replied to, in much the same way as toggling the "@" button. When `allowedMentions` is given, `replied_user` [defaults to `false`](https://discord.com/developers/docs/resources/channel#allowed-mentions-object), barring a bug in discord.js or inaccurate documentation.

src/commands/index.ts
AverageHelper: That's not how TS works.

Since the files are transpiled before runtime, file introspection for `.ts` or `.test.ts` or even `.d.ts` files in the commands directory will not work.

We \*could\* check for `.js` files, but we'd have no way of knowing at runtime whether those files are properly-formatted command definitions. To avoid having to handle potentially-catastrophic errors that may never happen, we import the file. This way, TypeScript has a chance to check that the commands we want are correctly structured, with the proper properties and `execute` method present. A direct import at the top of the file also informs code analyzers (and possible code bundlers, should we go that route) what files we use, so that those command files do not get stripped in a bundle or skipped in a security analysis as dead code.

AverageHelper:
I'm totally down to abstract those out! Would greatly simplify the unit tests for `main`.

IDK the best structure tho. I think the way to do this cleanly will be clearer when we have more events that need handling, so we have a better idea of the scope of logic we'd be moving. This PR is also huge as it is. Can we wait for the next PR to make that abstraction?

src/commands/index.ts
AverageHelper: TypeScript also ensures that the command actually registered, so long as we update the index. This simplifies unit tests down the road. Filesystem introspection and dynamic imports are both slow and error-prone, so I'd want lots of tests around those, and constructing them would be difficult.

src/@types/CommandContext.d.ts
AverageHelper: I'm not sure when to bring it up, tbh. It's not really an issue until we start running the bot full-time and need better logging and reporting tools. At the very least, there's code in here that's ready to take a new logger when/if we decide to do that.

AverageHelper: But yes, we should discuss this at some point, either now or later.

JstnMcBrd:
Sure, as long as we plan on doing it sometime. It isn't a high priority - everything works right now.

As for how to do it, I expect we could just use a similar format as your command handler. Make an `events` subfolder with an `index.ts`, have `main.ts` register all the events and then login, etc.

JstnMcBrd (approved):
All comments resolved 👍

src/helpers/guards/isNonEmptyArray.ts
```diff
@@ -0,0 +1,3 @@
+export function isNonEmptyArray<T>(array: Array<T>): array is NonEmptyArray<T> {
```
ZYancey, line 1: I'm going to be honest I don't super love the fact that we are writing code for this when an empty array could just be tested with a !!exampleArray as it would return false if the array is empty. I could be wrong and maybe this function serves a larger purpose but it does feel a little wrong.

ZYancey (approved):
Approved with comment

JstnMcBrd:
Hate to be a stickler, but there's also some unresolved code coverage warnings.

README.md
```diff
@@ -72,7 +72,7 @@ DISCORD_TOKEN=YOUR_TOKEN_GOES_HERE
 
 ### Invite your bot to your server
 
-Go to https://discordapi.com/permissions.html#377957215296 and paste in your bot's client ID to get an invite link.
+Go to https://discordapi.com/permissions.html#378091424832 and paste in your bot's client ID to get an invite link.
```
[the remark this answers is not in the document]
gmacgre: Once w have the basic permissions for command handling and other items, I don't think so.

src/helpers/guards/isNonEmptyArray.ts
AverageHelper: To be clear, you're absolutely right that this function shouldn't be necessary, but I've found it helpful in enough spots that I brought it in here.

README.md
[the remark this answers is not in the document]
AverageHelper: ^ what he said

src/@types/CommandContext.d.ts
```diff
@@ -0,0 +1,103 @@
...
+		readonly interaction: CommandInteraction;
+
+		/** Our own signed-in Discord client. */
+		readonly client: Client<true>;
+
+		/** The place to submit informative debug messages. */
+		readonly logger: Console; // TODO: More extensible logger than `console`?
+
+		/** The guild in which the command was invoked. */
+		readonly guild: Guild | null;
```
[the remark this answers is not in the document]
AverageHelper: Only that `Guild?` is not valid TypeScript syntax

src/commands/help.ts
```diff
@@ -0,0 +1,9 @@
+export const help: GlobalCommand = {
+	name: 'help',
+	description: 'Prints the list of commands',
+	requiresGuild: false,
+	async execute({ reply }) {
```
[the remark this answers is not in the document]
AverageHelper: Pretty much. Separating the event handling and execution layers this way helps to decouple the command's own logic from the details of how Discord's API works.

src/handleInteraction.ts
```diff
@@ -0,0 +1,192 @@
...
+		options: interaction.options.data,
+		logger,
+		async prepareForLongRunningTasks(ephemeral) {
+			try {
+				await interaction.deferReply({ ephemeral });
+			} catch (error) {
+				logger.error('Failed to defer reply to interaction:', error);
+			}
+		},
+		async replyPrivately(options, viaDM: boolean = false) {
```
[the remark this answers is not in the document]
JstnMcBrd: Personally I prefer to capitalize all parts of an abbreviation for clarity.

src/commands/help.ts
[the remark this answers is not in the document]
AverageHelper: And yes, every command will receive a `Context` object of this same shape. We can debate on the shape of that object as time goes on and more commands teach us what our specific needs are, but overall I intend to keep up this pattern.

src/commands/index.ts
```diff
@@ -0,0 +1,17 @@
+export const allCommands = new Map<string, Command>();
+
+/**  Install commands here:  **/
+import { help } from './help';
+_add(help);
+
+export function _add(cmd: Command): void {
```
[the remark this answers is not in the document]
AverageHelper: That's it. I export it literally just to test it.

src/handleInteraction.test.ts
```diff
@@ -0,0 +1,389 @@
...
+		test('sends deferrment when requested', async () => {
+			await expect(handleInteraction(interaction, mockConsole)).resolves.toBeUndefined();
+			expect(mockInteractionDeferReply).toHaveBeenCalledOnce();
+		});
+
+		test('edits interaction reply if deferred', async () => {
+			interaction = { ...interaction, deferred: true } as unknown as CommandInteraction;
+			await expect(handleInteraction(interaction, mockConsole)).resolves.toBeUndefined();
+			expect(mockInteractionEditReply).toHaveBeenCalled();
+			expect(mockInteractionEditReply).toHaveBeenCalledWith('yo');
```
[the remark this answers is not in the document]
AverageHelper: It is indeed. I'll remove it, and the other ones I spot

src/handleInteraction.test.ts
```diff
@@ -0,0 +1,389 @@
...
+			member: null,
+			guild: null,
+			channel: { type: ChannelType.DM },
+		} as unknown as CommandInteraction;
+
+		await expect(handleInteraction(interaction, mockConsole)).resolves.toBeUndefined();
+		expect(mockExecute).not.toHaveBeenCalled();
+	});
+
+	test('does nothing if the command is not found', async () => {
```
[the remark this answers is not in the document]
AverageHelper: As far as I know, Discord validates command requests against the list we give them. But if that list ever becomes out of date (we forget to deploy commands, or the deployment failed, or a few other cases) then some graceful failure is a good idea.

Right now, a log message gets sent after every command invocation indicating who sent it and what they sent. If the command is not valid, nothing more gets printed. We could change that with a `console.warn` though.

Ideas?

src/handleInteraction.test.ts
```diff
@@ -0,0 +1,389 @@
...
+			member: null,
+			channel: { type: ChannelType.DM },
+			reply: mockInteractionReply,
+		} as unknown as CommandInteraction;
+
+		await expect(handleInteraction(interaction, mockConsole)).resolves.toBeUndefined();
+		expect(mockExecute).not.toHaveBeenCalled();
+		expect(mockInteractionReply).toHaveBeenCalledOnce();
+		expect(mockInteractionReply).toHaveBeenCalledWith({
+			content: "Can't do that here",
```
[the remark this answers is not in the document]
AverageHelper: Great for a follow-up PR, yeah!

src/handleInteraction.ts
```diff
@@ -0,0 +1,192 @@
...
+	logger.debug(`User ${logUser(interaction.user)} sent command: '${interaction.commandName}'`);
+
+	const command = allCommands.get(interaction.commandName);
+	if (!command) return;
+
+	logger.debug(
+		`Calling command handler '${command.name}' with options ${JSON.stringify(
+			interaction.options,
+			undefined,
+			2
```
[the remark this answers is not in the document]
AverageHelper: Ohh, good point. Fixing that now.

src/handleInteraction.ts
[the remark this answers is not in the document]
AverageHelper: It is not, and I agree. Fixed.

src/handleInteraction.ts
```diff
@@ -0,0 +1,192 @@
...
+				// FIXME: We didn't actually send the reply if we errored out
+				logger.info(
+					`Sent ephemeral reply to User ${logUser(interaction.user)}: ${JSON.stringify(options)}`
+				);
+			}
+		},
+		async followUp(options) {
+			if (
+				typeof options !== 'string' &&
+				(!('reply' in options) || options.reply === false || options.reply === undefined) &&
```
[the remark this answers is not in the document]
AverageHelper: The `'reply' in options` bit tells TypeScript that it's safe to access `options.reply`. Since `options` can be neither `null` or `undefined`, the `?.` operator is useless to us here.

As for why we couldn't just use `!options.reply` there, that's due to an ESLint configuration I have checking for nullable values in conditionals, to make those checks explicit. Sometimes, we may want different behavior for nullish values vs empty or falsy ones, and I don't always keep the distinction in my mind when I'm writing conditions, hence the ESLint rule enforcing explicit logic.

src/@types/CommandContext.d.ts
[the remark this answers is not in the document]
AverageHelper: EDIT: I seem to have been mistaken. TIL that `Guild?` means `Guild | null`.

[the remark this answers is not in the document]
AverageHelper: I think (for now at least) I'd like to leave `| null` suffixes, so that my monkey brain down the line doesn't confuse `Foo?` for `Foo | undefined`, when it really means `Foo | null`.

ZYancey (approved):

src/handleInteraction.test.ts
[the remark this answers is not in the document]
Plyb: One thing Ze Kaiser did is that it would send errors to a particular channel on a guild if something was broken, as well as to admins. I don't think we need it to DM admins, but maybe implementing the error channel for this would be a good idea? I don't think we need to do it in this PR since it's already big enough, but what are your thoughts on that solution?

src/handleInteraction.ts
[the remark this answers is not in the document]
Plyb: sounds good

src/helpers/guards/isError.test.ts
```diff
@@ -0,0 +1,45 @@
...
+	test('returns true for Error subclasses', () => {
+		expect(isError(new EvalError('This is an error'))).toBeTrue();
+		expect(isError(new RangeError('This is an error'))).toBeTrue();
+		expect(isError(new ReferenceError('This is an error'))).toBeTrue();
+		expect(isError(new SyntaxError('This is an error'))).toBeTrue();
+		expect(isError(new TestError())).toBeTrue();
+		expect(isError(new TypeError('This is an error'))).toBeTrue();
+	});
+
+	test.each`
```
Plyb, line 32: the h*cc is this syntax?

src/helpers/guards/isError.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * Checks whether a value is a NodeJS error.
+ *
+ * The normal `instanceof` and `isNativeError` checks don't
+ * work, because the `Error` type alone doesn't include
+ * certain common error properties. Since those properties
+ * are optional anyway, this function coerces `Error` to
+ * `NodeJS.ErrnoException`.
+ */
+export function isError(tbd: unknown): tbd is NodeJS.ErrnoException {
```
Plyb, line 10: neat syntax on the type

src/main.test.ts
```diff
@@ -1,10 +1,192 @@
...
+			deploy: false,
+			revoke: false,
+		});
+	});
+
+	afterEach(() => {
+		jest.restoreAllMocks();
+	});
+
+	test('disables @everyone pings', async () => {
```
Plyb, line 100: lol nice, getting that out of the way quickly

src/main.ts
```diff
@@ -1,7 +1,91 @@
...
+	});
+
+	const args = parseArgs();
+
+	console.info('*Yawn* Good morning!');
+	console.info(`Starting CS Bot v${version}...`);
+	console.debug(`Node ${process.version}`);
+
+	client.on('ready', async client => {
+		// If we're only here to revoke commands, do that and then exit
```
Plyb, line 36: typo in the comment here "revoke" > "deploy"

Plyb (changes requested):
My only other comment is that we should make tickets for the TODOs in here before/once this is merged

src/handleInteraction.test.ts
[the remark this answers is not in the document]
AverageHelper: Sounds good to me, as long as we don't need to hard-code the error channel here. Env vars for the win, lol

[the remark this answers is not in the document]
AverageHelper: I'll make a TODO on it

src/helpers/guards/isError.test.ts
AverageHelper: This is a [Template Literal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals). They can be used as function arguments like this to do really weird JavaScript magic. In this case, Jest uses this as their first-party solution for parameterized tests. I hate it, but it's well-documented Jest, and it works, lol

src/helpers/guards/isError.ts
AverageHelper: ikr

src/handleInteraction.ts
```diff
@@ -0,0 +1,193 @@
...
+				return false;
+			}
+		},
+		sendTyping() {
+			void interaction.channel?.sendTyping();
+			logger.debug(
+				`Typing in channel ${interaction.channel?.id ?? 'nowhere'} due to Context.sendTyping`
+			);
+		},
+	};
```
JstnMcBrd, line 174: This context object declaration is really lengthy. It sorta distracts from the purpose of `handleInteraction` because most of the logic is not even used here. Plus, it gets generated every single time an interaction is received.

Is there any way to move this logic out of this file and have it be loaded in as an object (and edited) at runtime? It would really clean up this file. And then we can edit the command context whenever necessary without messing with this event handler.

AverageHelper: Good idea. I think each of the methods could be abstracted into separate files, sortof like the `replyPrivately` and `sendMessageInChannel` functions are. Those two helpers could probably be combined with the context methods somewhat.

I think constructing the context is part of the purview of "handling the interaction," as part of that job is to select the appropriate command and execute it, which necessarily involves a context specific to the invocation. The methods tho could definitely be defined with factory functions elsewhere, since they only need the current interaction and the logger instance to know what to do.

That work could be done in a separate PR tho. Once this PR is in, then the team can work on various parts of the bot, including commands and even the context construction. This PR mainly lays down the initial bones and interface joints for the whole thing.

AverageHelper: IMO length itself is not a huge concern, but when the long function starts dealing with things that it should not need to deal with, like the ins and outs of Discord's Interactions API, that's not ideal.

I think factory functions might be the way to go here, tho regardless of how we solve this problem, I think the solution can be non-breaking enough to warrant a separate PR.

AverageHelper: I've created #21 to dive into this further, if this needs more discussion.

Plyb (approved):

PR #23, opened 2022-09-29
Basic `/help` functionality

### LINKED ISSUE #8: Slash Command - basic Help

AverageHelper:
At its very basic, `/help` lists the commands available to the user. This is at least a start.

Screen Shot 2022-09-28 at 23 23 19

This command will automatically include new commands as they're added in `src/commands/index.ts`.

Closes #8

### DIFF
--- a/src/commands/__snapshots__/help.test.ts.snap
+++ b/src/commands/__snapshots__/help.test.ts.snap
@@ -1,0 +1,25 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`help presents an ephemeral embed with all available global and guild-bound commands 1`] = `
+Object {
+  "fields": Array [
+    Object {
+      "name": "\`/help\`",
+      "value": "Prints the list of commands",
+    },
+  ],
+  "title": "All commands",
+}
+`;
+
+exports[`help presents an ephemeral embed with all available global commands 1`] = `
+Object {
+  "fields": Array [
+    Object {
+      "name": "\`/help\`",
+      "value": "Prints the list of commands",
+    },
+  ],
+  "title": "All commands",
+}
+`;
--- a/src/commands/help.ts
+++ b/src/commands/help.ts
@@ -1,9 +1,43 @@
+import { EmbedBuilder } from 'discord.js';
+
 export const help: GlobalCommand = {
 	name: 'help',
 	description: 'Prints the list of commands',
 	requiresGuild: false,
-	async execute({ reply }) {
-		// TODO: finish this command
-		await reply('Hello, world!');
+	async execute({ guild, reply }) {
+		// Dynamic import here b/c ./index depends on this file
+		const { allCommands } = await import('./index');
+
+		const embed = new EmbedBuilder() //
+			.setTitle('All commands');
+
+		function embedCommand(command: Command): void {
+			embed.addFields({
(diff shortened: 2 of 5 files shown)


src/commands/help.ts
```diff
@@ -1,9 +1,43 @@
...
+		}
+
+		for (const command of allCommands.values()) {
+			if (guild) {
+				// We're in a guild. We should check the user's permissions,
+				// and skip this command if they aren't allowed to use it.
+				// TODO: Grab the guild's configured permissions for this channel
+				embedCommand(command);
+				continue;
+			}
```
gmacgre, line 28: Is this TODO being worked on at the moment?

src/commands/help.ts
```diff
@@ -1,9 +1,43 @@
...
+
+			// We're in DMs; command should have dmPermission if we're to print it.
+			// Discord defaults dmPermission to `true`, so `undefined` should behave that way.
+			// See https://discordjs.guide/interactions/slash-commands.html#dm-permission
+			if (command.dmPermission === false) continue;
+
+			embedCommand(command);
```
gmacgre, line 35: Looks like this section is not covered in tests. We should probably make sure that it works in DMs as well before merging.

gmacgre (changes requested):
Other than those two small fixes, LGTM

src/commands/help.ts
gmacgre: Or is the building of perms checking something that we should add to the docket and circle back to?

AverageHelper: Might be complicated. I poked at the problem a bit, and I realized I need to learn how this new guild-centric permission system works. I'm used to the bot maintaining a database of its own command permissions, but slash commands don't do that anymore, so I need to re-learn how the APIs relate to the UI.

If someone can pick that up before I get around to it, be my guest. Meantime, since Discord enforces command permissions anyway, I'm fine pushing this one down the road.

gmacgre: Alright. I think we can revisit this when we get a little more experience with the API then. I'll add is as a ticket for future building.

src/commands/help.ts
AverageHelper: Good point. Adding a test now

gmacgre (approved):
LGTM. I think we have everything we need now for this command.

gmacgre:
Note- We may want to wait for one more person to approve this before merging.

src/commands/__snapshots__/help.test.ts.snap
```diff
@@ -0,0 +1,25 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
JstnMcBrd, line 1: What's this extra file all about? And the weird `__snapshots__` subdirectory? Clutter?

AverageHelper: That's a smapshot file, generated by Jest the first time a snapshot is taken. Future tests will compare outputs against this file, and yell if things change unexpectedly, indicating a failed test.

JstnMcBrd (approved):

JstnMcBrd:
Oh I forgot to mention, probably should keep the changelog up to date.

AverageHelper:
Fair point

PR #33, opened 2022-10-01
Added Profile Slash Command to Bot

ZYancey:
This functionality replicates the !avatar command of Ze Kaiser, utilizing the new framework for slash commands


CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 0.1.4 - 2022-10-01
```
AverageHelper, line 8: Be sure to update package.json too

AverageHelper (changes requested):

src/commands/profile.ts
```diff
@@ -0,0 +1,34 @@
+import { ApplicationCommandOptionType } from 'discord.js';
+
+export const profile: GlobalCommand = {
+	name: 'profile',
+	description: 'Responds with the Profile Picture of the mentioned user',
+	options: [
+		{
+			name: 'user',
+			description: 'The user to get the profile picture of',
+			type: ApplicationCommandOptionType?.User,
```
AverageHelper, line 10: `ApplicationCommandOptionType` isn't nullish, so we don't need to coalesce here.

```suggestion
			type: ApplicationCommandOptionType.User,
```

src/commands/profile.ts
```diff
@@ -0,0 +1,34 @@
...
+		},
+	],
+	requiresGuild: false,
+	async execute({ interaction, options }) {
+		const user = interaction.user;
```
AverageHelper, line 15: The context includes the `user` property of the interaction, and a `reply` helper that handles errors and certain `CommandInteraction` edge cases.

```suggestion
	async execute({ user, options, reply }) {
```

src/commands/profile.ts
```diff
@@ -0,0 +1,34 @@
...
+		const param = options[0];
+		if (param?.value === undefined) {
+			await interaction.reply({
+				content: user.avatarURL({ extension: 'png', size: 2048 }) as string,
```
AverageHelper, line 19: Rather than force-casting to `string`, we should probably handle the case where this function returns `null`, perhaps by replying with some error message about the caller not having an avatar.

src/commands/profile.ts
```diff
@@ -0,0 +1,34 @@
...
+				ephemeral: true,
+			});
+		} else if (param?.user?.avatar !== null) {
+			await interaction.reply({
+				content: param?.user?.avatarURL({ extension: 'png', size: 2048 }) as string,
```
AverageHelper, line 24: Since the previous condition handled `param` being `undefined`, these two cases don't need to coalesce.
```suggestion
		} else if (param.user?.avatar !== null) {
			await interaction.reply({
				content: param.user?.avatarURL({ extension: 'png', size: 2048 }) as string,
```

AverageHelper: We should maybe also handle this `null` case, rather than force-casting it away.

AverageHelper (changes requested):

src/main.test.ts
```diff
@@ -183,7 +183,7 @@ describe('main', () => {
 	});
 
 	test('reports interaction errors', async () => {
-		const interactionError = new Error('Failed to handle ineracion. This is a test.');
+		const interactionError = new Error('Failed to handle ineraction. This is a test.');
```
JstnMcBrd, line 186:
```suggestion
		const interactionError = new Error('Failed to handle interaction. This is a test.');
```
Almost fixed the spelling error, just one more 't'.

src/commands/profile.ts
AverageHelper: The reason this failed the unit tests before is that we mock `discord.js` in `main.test.ts`. There are some ways to fix that:
- Import the actual `ApplicationCommandOptionType` object from `discord.js` and re-export that in the mock.
- Mock command/index.ts in `main.test.ts`. This may not work, depending on how deep that import is in the dependency tree.

CHANGELOG.md
gmacgre, line 8: Hate to be a stickler, but shouldn't this be 0.2.0, since it's more adding functionality to the bot as opposed to a patch? Avg might help draw the line here though.

src/commands/profile.ts
```diff
@@ -0,0 +1,34 @@
...
+				ephemeral: false,
+			});
+		} else {
+			await interaction.reply({
+				content: "This user doesn't have an avatar",
+				ephemeral: true,
+			});
+		}
```
gmacgre, line 32: If we're going to take in the suggestion from Avg on checking that a [URL exists](https://github.com/BYU-CS-Discord/CSBot/pull/33/files#r985131421), this might work better as a function that we can call whenever there is an error getting the avatar at any stage of the process.

gmacgre (changes requested):

CHANGELOG.md
AverageHelper: ¯\_(ツ)_/¯

Makes sense to me

src/commands/profile.ts
ZYancey: Was able to get around this by simplifying the code considerably to remove duplicated calls

src/commands/profile.ts
```diff
@@ -0,0 +1,35 @@
...
+		const param = options[0];
+		const target = param ? param.user : user;
+		// if the user is not null, then we will use that user, otherwise we will use the user that invoked the command.
+
+		if (target) {
+			await reply({
+				content:
+					target.avatarURL({ extension: 'png', size: 2048 }) ??
+					"Something went wrong, This user doesn't have an avatar!",
+				ephemeral: target.avatarURL() === null,
```
AverageHelper, line 24: Calling `avatarURL()` again may have a different result from the first call, so best call it only once:
```suggestion
			const content = target.avatarURL({ extension: 'png', size: 2048 });
			await reply({
				content: content ?? "Something went wrong, This user doesn't have an avatar!",
				ephemeral: content === null,
```

AverageHelper (changes requested):
Looks good! Could you maybe add some unit tests too? Just something to make sure that `avatarURL` gets called, and another to make sure that some fallback message gets displayed if something goes wrong

gmacgre (dismissed):

Plyb (dismissed):
With Avg's requested change, looks good otherwise

AverageHelper (approved):

JstnMcBrd (approved):

AverageHelper (approved):

PR #36, opened 2022-10-01
Logger abstraction

JstnMcBrd:
The logger was already partially abstracted before, but it required passing logger around through parameters in methods. This was causing problems for event handler abstraction.

Now the logger is in a separate file called `logger.ts` that contains a method that returns the logger to use (the console, for now). We use a method instead of a constant to make mocking easier for tests.

All files that used the logger before now pull the logger from `logger.ts` instead of passing through parameters. I also updated tests to reflect this fact by mocking `logger.ts`.

### DIFF
--- a/src/logger.ts
+++ b/src/logger.ts
@@ -1,6 +1,6 @@
 /**
- * A method that returns the default logger for this application
- * Uses a function instead of a constant to allow mocking
+ * A method that returns the default logger for this application.
+ * Uses a function instead of a constant to allow mocking.
  * @returns The default logger for this application
  */
 export function getLogger(): Console {
--- a/src/logger.ts
+++ b/src/logger.ts
@@ -1,0 +1,8 @@
+/**
+ * A method that returns the default logger for this application
+ * Uses a function instead of a constant to allow mocking
+ * @returns The default logger for this application
+ */
+export function getLogger(): Console {
+	return console;
+}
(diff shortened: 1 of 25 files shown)


src/logger.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * A method that returns the default logger for this application
+ * Uses a function instead of a constant to allow mocking
```
AverageHelper, line 3: Just so this doesn't look like a run-on sentence in VS Code's popover view:

```suggestion
 * A method that returns the default logger for this application.
 * Uses a function instead of a constant to allow mocking.
```

AverageHelper (changes requested):

src/logger.ts
JstnMcBrd: It's not a complete sentence though. My high school English teachers would be mortified!

AverageHelper (dismissed):

src/logger.ts
AverageHelper: You could always make them complete sentences, lol

JstnMcBrd:
I forgot some parts. I'll be pushing more changes soon.

AverageHelper (approved):

gmacgre (dismissed):
Making a personal note here- we probably should have merged this in before `xkcd`. I'll need to go back and make sure that the logging there is up to date.

JstnMcBrd:
Since you just merged main into this branch, I can go and edit xkcd's code before merging.

gmacgre (approved):

AverageHelper (approved):

PR #38, opened 2022-10-02
Event Handler Abstraction

### LINKED ISSUE #28: Abstract out the event handling system
See [this comment](https://github.com/BYU-CS-Discord/CSBot/pull/14#issuecomment-1260551210) under PR#14

JstnMcBrd:
This is a chonky pull request. Lots of new files, and lots of little edits all over the place.

- Added a new type for Event Handlers
- Added a `./events` folder with an `index.ts` that gathers all event files, similar to `./commands`
- Moved all event handlers out of `main.ts` and into individual event files
- Moved all event handler testing code out of `main.test.ts` into individual test files
- Moved command-handling code from `handleInteraction.ts` into the `interactionCreate` event file
- Updated all affected test files
- Wrote new tests for all new code - 100% coverage on all additions
- On the side, I made `xkcd.test.ts` mock the logger file like the other tests

There's a lot here, so I'm sure it's a little messy. Don't be shy with requesting changes.

Closes #28


src/@types/EventHandler.d.ts
```diff
@@ -0,0 +1,25 @@
+// IDK why this line is necessary, but if I remove it, I can't use EventHandler anywhere
+import type { Awaitable } from 'discord.js';
+
+declare global {
+	/**
+	 * Defines an event handler.
+	 */
+	interface EventHandler {
```
AverageHelper, line 8: I think we can make this interface loads safer by making it generic over the keys of the `ClientEvents` interface from discord.js. We can then augment the `name` and `execute` properties according to the event name. This way, event handlers always know exactly what data to expect from the API

AverageHelper (changes requested):
This is the first file I've looked at, but I think you might like working on this one:

src/@types/EventHandler.d.ts
JstnMcBrd: Could we save that for a future PR? I'm still learning TypeScript and it's been mostly trial-and-error so far. That's probably better left to someone who knows what their doing.

AverageHelper (changes requested):

src/@types/EventHandler.d.ts
AverageHelper: I'll get right on it once this is merged

src/events/index.ts
```diff
@@ -0,0 +1,57 @@
...
+
+// Internal dependencies
+import { getLogger } from '../logger';
+const logger = getLogger();
+
+/**
+ * The list of all event handlers. DO NOT EDIT DIRECTLY.
+ * registerEventHandlers() will pull from this list.
+ */
+export const allEventHandlers = new Map<string, EventHandler>();
```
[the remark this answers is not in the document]
JstnMcBrd: I'll get on that.

src/events/interactionCreate.test.ts
```diff
@@ -0,0 +1,217 @@
...
+import { ChannelType } from 'discord.js';
+
+// Mock allCommands to isolate our test code
+const mockAllCommands = new Map<string, Command>();
+jest.mock('../commands', () => ({
+	allCommands: mockAllCommands,
+}));
+
+// Create two mock commands to track handler behavior
+const mockGlobalExecute = jest.fn();
```
Plyb, line 12: To make sure I'm understanding, is interactionCreate an event specifically dealing with commands?

Plyb (dismissed):

src/events/interactionCreate.test.ts
AverageHelper: Yep. [Here](https://discordjs.guide/creating-your-bot/event-handling.html#event-handling) are some docs on that

src/@types/EventHandler.d.ts
```diff
@@ -0,0 +1,24 @@
...
+		/**
+		 * The name of the event this handler is for
+		 */
+		name: string;
```
AverageHelper, line 12: Could we make this `keyof ClientEvents`? The autocomplete buff would be nice for writing new event handlers, even if the `execute` method still uses `any` params for now

AverageHelper (changes requested):

src/events/ready.test.ts
```diff
@@ -0,0 +1,123 @@
...
+			revoke: true,
+		});
+		await expect(ready.execute(client)).resolves.toBeUndefined();
+		expect(mockDeployCommands).toHaveBeenCalledWith(new MockClient());
+		expect(mockRevokeCommands).not.toHaveBeenCalled();
+	});
+
+	test('verifies command deployments', async () => {
+		await expect(ready.execute(client)).resolves.toBeUndefined();
+		expect(mockVerifyCommandDeployments).toHaveBeenCalledWith(new MockClient());
```
AverageHelper, line 116: Shouldn't these reference the same client instance?
```suggestion
		expect(mockVerifyCommandDeployments).toHaveBeenCalledWith(client);
```

This might work anyway as-is because Jest is comparing the properties of the objects, rather than using a strict identity comparison, but that's an implementation detail that might change, based on the verbiage of the assertion method. (IMO, "have been called with" implies strict identity in the case of non-scalar values.)

AverageHelper (changes requested):

src/commands/xkcd.test.ts
```diff
@@ -1,9 +1,20 @@
 import axios from 'axios';
-import { xkcd } from './xkcd';
 jest.mock('axios');
 
 const mockedAxios = axios as jest.Mocked<typeof axios>;
 
+jest.mock('../logger');
+import { getLogger } from '../logger';
+const mockGetLogger = getLogger as jest.Mock;
+mockGetLogger.mockImplementation(() => {
```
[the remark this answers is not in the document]
JstnMcBrd: Add an issue for it so we don't forget.

src/@types/EventHandler.d.ts
JstnMcBrd: I suggest making an issue for it before we forget.

src/events/ready.test.ts
JstnMcBrd: Oh you're right, I meant to make that change after figuring out how to mock only the Client, but I forgot a few spots apparently.

src/commands/index.ts
```diff
@@ -1,19 +1,35 @@
...
-_add(xkcd);
+/**
+ * A read-only list of all commands.
+ * @public
+ */
+export const allCommands: ReadonlyMap<string, Command> = _allCommands;
 
+/**
+ * Adds an command to the list of all commands.
+ * Only public for testing purposes. Avoid using.
```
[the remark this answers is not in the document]
JstnMcBrd: I figure it isn't really important enough to have a strict standard like that. I just prefer to be extremely explicit in my comments, so I'll likely put that on all of my changes. Others don't have to do the same.

And you're right, the wording is weird. I'll adjust that.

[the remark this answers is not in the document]
AverageHelper: Didn't ask, but my two cents: I would hope that the leading `_` is odd enough that most devs would think twice about importing and using such structures in production code. At least, I'd hope that enough people on the team that reviews PRs would know to catch those and inform devs who aren't yet aware that the leading underscore is not commonly "ok" to just import and use.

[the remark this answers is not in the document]
AverageHelper: That said, it's entirely possible that this is the C/Objective-C/Swift dev in me that thinks of the leading underscore as a "don't touch unless you know what you're doing" indicator.

src/@types/EventHandler.d.ts
JstnMcBrd: I tried this out and it mostly worked - however, it did break the test case where I give the index fake event handlers to make sure it's registering them properly. I'm trying to figure out a workaround.

src/events/ready.ts
```diff
@@ -0,0 +1,57 @@
+// Dependencies
+import type { Client, ClientPresence } from 'discord.js';
+import { ActivityType, ClientEvents } from 'discord.js';
```
AverageHelper, line 3: We don't use `ClientEvents` in this file

AverageHelper (approved):

src/events/ready.ts
AverageHelper: "Approved" anyway because I hit the wrong button, lol

ZYancey (approved):

JstnMcBrd:
I'm taking this opportunity to abstract out the logger mocking. Wait to merge until I finish.

AverageHelper (approved):

AverageHelper:
You could also merge now, and abstract out the logger mock in another PR :P

JstnMcBrd:
You're right, I'll do that. Keep separate issues separate.

PR #39, opened 2022-10-04
Logger mocking abstraction

JstnMcBrd:
Plyb made a good point in [this comment](https://github.com/BYU-CS-Discord/CSBot/pull/38#discussion_r985362920) that mocking the default logger was something nearly every test file had to do, and that was causing significant code duplication.

So I created a `mockLogger.ts` file that mocks the logger automatically, and then made all test scripts import the mock they need to test output.

Unfortunately, naming the mock logger file `mockLogger.test.ts` caused jest to throw errors because it expected the file to have tests inside. So I had to name it `mockLogger.ts` instead, which made ESLint throw errors because it didn't like the usage of jest and other testing syntax. So I had to add rules to the configs for TypeScript and ESLint to treat `mockLogger.ts` as a test file.

Besides that, everything else went smoothly.


src/helpers/testing/mockLogger.ts
```diff
@@ -0,0 +1,51 @@
+/**
```
AverageHelper, line 1: Jest has [this neat little thing](https://jestjs.io/docs/manual-mocks#mocking-user-modules) where you can save a mock version of a module in a subfolder called `__mocks__`. For example, to mock `foo.ts`, you'd create `__mocks__/foo.ts`. ESLint is already configured to treat these mocks as Jest components, and TypeScript is already configured to omit these mocks from the build output like other test files.

To consume the mock in a test, each test needs only run `jest.mock("path/to/foo")`, and Jest automatically replaces `foo` with the mocked version in the adjacent `__mocks__` directory.

AverageHelper (changes requested):

.eslintrc
```diff
@@ -101,7 +101,8 @@
 			},
 			"rules": {
 				"prettier/prettier": "warn",
-				"max-nested-callbacks": "off" // unit tests involve a lot of nested callbacks
+				"max-nested-callbacks": "off", // unit tests involve a lot of nested callbacks
+				"@typescript-eslint/unbound-method": "off" // unit tests also involve passing unbounded methods
```
AverageHelper, line 105: I disagree. Messing with method binding can cause JS-land `this`-binding weirdness that I'd rather ESLint catch by default. Typically, if ESLint throws this at you, you should redesign to either call the method in a lambda, or build a better method. If you _must_ hand around an unbound method, you shouldn't.

But if for sure you really promise it won't break anything ever, then throw an `eslint-ignore-next-line` where needed to silence the warning at the use site.

.eslintrc
```diff
@@ -90,7 +90,7 @@
 
 	"overrides": [
 		{
-			"files": ["src/**/*.test.ts", "src/**/__mocks__/**/*.ts"],
+			"files": ["src/**/*.test.ts", "src/**/__mocks__/**/*.ts", "src/**/testing/**/*.ts"],
```
AverageHelper, line 93: Unneeded if the mock was in a `__mocks__` folder.

src/commandContext/sendTyping.test.ts
```diff
@@ -1,14 +1,7 @@
...
-import { getLogger } from '../logger';
-const mockGetLogger = getLogger as jest.Mock;
-const mockConsoleDebug = jest.fn();
-mockGetLogger.mockImplementation(() => {
-	return {
-		debug: mockConsoleDebug,
-	} as unknown as Console;
-});
+// Mock the logger so nothing is printed
+import {} from '../helpers/testing/mockLogger';
```
AverageHelper, line 4: This is the typical syntax for a side-effect-only import:

```suggestion
import '../helpers/testing/mockLogger';
```

src/commands/xkcd.test.ts
```diff
@@ -3,17 +3,8 @@ jest.mock('axios');
...
-mockGetLogger.mockImplementation(() => {
-	return {
-		debug: () => undefined,
-		info: () => undefined,
-		warn: () => undefined,
-		error: () => undefined,
-	} as unknown as Console;
-});
+// Mock the logger so nothing is printed
+import {} from '../helpers/testing/mockLogger';
```
AverageHelper, line 7:
```suggestion
import '../helpers/testing/mockLogger';
```

src/events/index.test.ts
```diff
@@ -20,15 +20,8 @@ jest.mock('discord.js', () => ({
...
-jest.mock('../logger');
-import { getLogger } from '../logger';
-const mockGetLogger = getLogger as jest.Mock;
-mockGetLogger.mockImplementation(() => {
-	return {
-		info: () => undefined,
-	} as unknown as Console;
-});
+// Mock the logger so nothing is printed
+import {} from '../helpers/testing/mockLogger';
```
AverageHelper, line 24:
```suggestion
import '../helpers/testing/mockLogger';
```

src/events/ready.test.ts
```diff
@@ -42,15 +42,8 @@ jest.mock('../helpers/actions/verifyCommandDeployments');
...
-jest.mock('../logger');
-import { getLogger } from '../logger';
-const mockGetLogger = getLogger as jest.Mock;
-mockGetLogger.mockImplementation(() => {
-	return {
-		info: () => undefined,
-	} as unknown as Console;
-});
+// Mock the logger so nothing is printed
+import {} from '../helpers/testing/mockLogger';
```
AverageHelper, line 46:
```suggestion
import '../helpers/testing/mockLogger';
```

src/helpers/actions/deployCommands.test.ts
```diff
@@ -1,17 +1,8 @@
...
-mockGetLogger.mockImplementation(() => {
-	return {
-		debug: () => undefined,
-		info: () => undefined,
-		warn: () => undefined,
-		error: () => undefined,
-	} as unknown as Console;
-});
+// Mock the logger so nothing is printed
+import {} from '../testing/mockLogger';
```
AverageHelper, line 5:
```suggestion
import '../testing/mockLogger';
```

src/helpers/actions/messages/replyToMessage.test.ts
```diff
@@ -1,14 +1,7 @@
...
-import { getLogger } from '../../../logger';
-const mockGetLogger = getLogger as jest.Mock;
-mockGetLogger.mockImplementation(() => {
-	return {
-		info: () => undefined,
-		error: () => undefined,
-	} as unknown as Console;
-});
+// Mock the logger so nothing is printed
+import {} from '../../testing/mockLogger';
```
AverageHelper, line 4:
```suggestion
import '../../testing/mockLogger';
```

src/helpers/actions/revokeCommands.test.ts
```diff
@@ -1,16 +1,7 @@
...
-mockGetLogger.mockImplementation(() => {
-	return {
-		debug: () => undefined,
-		info: () => undefined,
-		warn: () => undefined,
-		error: () => undefined,
-	} as unknown as Console;
-});
+// Mock the logger so nothing is printed
+import {} from '../testing/mockLogger';
```
AverageHelper, line 4:
```suggestion
import '../testing/mockLogger';
```

src/helpers/testing/mockLogger.ts
```diff
@@ -0,0 +1,51 @@
...
+
+jest.mock('../../logger');
+import { getLogger } from '../../logger';
+const mockGetLogger = getLogger as jest.Mock;
+
+const mockLogger = {
+	debug: jest.fn(),
+	info: jest.fn(),
+	warn: jest.fn(),
+	error: jest.fn(),
```
AverageHelper, line 19: Since you reference these down below in a way that TS thinks is referencing unbound methods, it might be best to declare and export these separately, and then construct `mockLogger`.

tsconfig.prod.json
```diff
@@ -1,5 +1,5 @@
 {
 	"extends": "./tsconfig.base.json",
 	"include": ["src/**/*.ts"],
-	"exclude": ["src/**/*.test.ts", "**/__mocks__/**/*.ts"]
+	"exclude": ["src/**/*.test.ts", "**/__mocks__/**/*.ts", "src/**/testing/**/*.ts"]
```
AverageHelper, line 4: Unnecessary if you use that `__mocks__` thingthing.

AverageHelper (changes requested):

src/helpers/testing/mockLogger.ts
JstnMcBrd: That is neat. I've been messing around with it, and it looks like if I use a `__mock__` file, I'll have to use unbound methods in test files unless I rewrite the logger system. The current implementation is rather inflexible with exports.

I've converted this PR into a draft in the meantime, because it might take me a while.

JstnMcBrd: Actually, I'm convinced this is impossible without using unbounded methods somewhere.

AverageHelper: Need to hop on a call to talk it over?

JstnMcBrd: Never mind, I figured out a way. It's just a little hacked. I'll upload soon and we'll see what you think.

src/logger.ts
```diff
@@ -1,8 +1,56 @@
...
  */
-export function getLogger(): Console {
-	return console;
+const logger: Console = console;
+export default logger;
+
+/**
+ * A wrapper method for the `debug` endpoint of the default logger.
+ * In most situations, it wouldn't make sense to import a single logger endpoint,
+ * but separate exports are necessary for creating jest mocks for testing.
```
AverageHelper, line 13: I wouldn't say it's necessary, only that they make the test code nicer, only having to import the function directly rather than importing the logger, cast methods of that logger to `jest.Mock`, and using that in the tests.

For conciseness' sake, I think these exports are a reasonable middle ground. Good job!

src/logger.ts
```diff
@@ -1,8 +1,56 @@
 /**
- * A method that returns the default logger for this application.
- * Uses a function instead of a constant to allow mocking.
- * @returns The default logger for this application
+ * The default logger for this application.
+ * Can be set to any object that has `debug`, `info`, `warn`, and `error` methods,
+ * as long as it can be cast to a `Console`.
+ * @public
```
AverageHelper, line 5: Doesn't the later `export` imply `@public`?

What do we mean by "public," anyway? This project isn't an API or SDK, the object is only ever used internally

AverageHelper (changes requested):

src/__mocks__/logger.ts
```diff
@@ -0,0 +1,41 @@
+'use strict';
```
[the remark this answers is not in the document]
AverageHelper: If so, then perhaps I've misconfigured `ts-jest`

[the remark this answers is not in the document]
JstnMcBrd: No, it was just suggested by the jest documentation. I'll remove it.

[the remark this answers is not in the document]
AverageHelper: It makes sense in JS land, which is why you'll see it in the compiled output .js files. TS adds it automatically at compile time

src/logger.ts
JstnMcBrd: It's more for consistency's sake. I like verbose documentation, and some modules have a mixed of internally-used methods and exported methods - and some have methods that are technically private but exported for testing reasons. I just like adding the tag to the code I write.

AverageHelper: Fair enough, carry on

src/logger.ts
```diff
@@ -1,8 +1,56 @@
...
  */
-export function getLogger(): Console {
-	return console;
+const logger: Console = console;
+export default logger;
```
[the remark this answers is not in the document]
JstnMcBrd: Considering the entire module is for the logger, and all sub-exports are sub-methods of the logger, it makes more sense to me to have the default export (IE, importing the entire module) simply be the logger.

It's similar to how discord.js works. You can import the entire module as the Discord object (`import Discord from 'discord.js'`), or import sub-classes of Discord (`import { Client } from 'discord.js'`).

It wouldn't make sense to have the `logger` object be on the same export level as its methods (`export { logger, debug, info, warn, error }`) when the methods belong to the logger (`logger.debug, logger.info, ...`). Can you see importing the entire module as the logger object makes more sense, when the logger object IS the module?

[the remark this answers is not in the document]
AverageHelper: We aren't exporting the entire module here though, we're exporting an object named `default`, which JS lends some syntactic sugar to let importers name the import whatever they want, transparently. TypeScript does some translation work to facilitate that import in the compiled output, (which you can read in `dist/main.js`), which work TS omits as unnecessary with named exports.

Discord.js doesn't use `export default`, at least not at the top level. They use named exports, in the CommonJS `exports.name` syntax, the same syntax to which TS translates our named imports. TS translates our import into a CommonJS `require` in the same way that it would translate an `import * as package from 'package'`. You can observe this by adding the following to `main.ts`, running `npm run build:only`, and observing `dist/main.js`:

```ts
import * as _Discord from 'discord.js';
import Discord from 'discord.js';
```

To really "import the entire module," importers should, IMO, use `import * as logger from './logger'` syntax. I'd prefer even that to default export, whose implicit renaming creates hidden room for inconsistent spellings and partial refactors, where the `as` operator or the `import { name }` syntax make clear that a rename is going on.

There's nothing I can do to prevent us using packages that rely on `export default`, that's their decision to make and our code can interop with that just fine. However, for the reasons given in [this article](https://basarat.gitbook.io/typescript/main-1/defaultisbad) (esp. good-practice CommonJS interop, consistent autocomplete, and auto-import), I plan to update our ESLint config to enforce the use of named exports in our code.

I agree that these parts shouldn't be on the same export level. Perhaps there can be some compromise? I'll approve the code as-is, but if it's merged, I intend to try a few things later to factor out that default export.

src/logger.ts
```diff
@@ -1,8 +1,56 @@
...
+
+/**
+ * A wrapper method for the `debug` endpoint of the default logger.
+ * In most situations, it wouldn't make sense to import a single logger endpoint,
+ * but separate exports are necessary for creating jest mocks for testing.
+ * @param message the message to print to `debug`
+ * @param optionalParams any additional messages or objects to print to `debug`
+ * @public
+ */
+export function debug(message?: never, ...optionalParams: Array<never>): void {
```
AverageHelper, line 18: Why `never`?

src/logger.ts
AverageHelper: That's fine at runtime if `logger.debug` is a `jest.Mock` object, because that value doesn't need access to `logger` using a bound `this`. TS doesn't know at compile time that it's importing a mock, hence the need to cast to `jest.Mock`. Verbose, but it gets the job done in cases where importing the function directly isn't doable.

src/logger.ts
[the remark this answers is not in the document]
AverageHelper: (I realize now that my comment has "If I may interject for a moment" vibes, but about the merits of named vs. `default` exports, lol)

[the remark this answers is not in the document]
AverageHelper: TBH, I mostly don't want us falling into the habit of using `export default` everywhere. I'd be plenty happy for now if you followed this SO answer to disallow `export default` by default, then add an `eslint-disable-next-line` marker in `logger.ts` to whitelist its use here.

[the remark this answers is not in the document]
JstnMcBrd: Ah, my bad, it appears I didn't understand TS exports as well as I thought.

Is there an alternative solution to organize the exports of `logger.ts` better (without using unbounded methods)?

src/logger.ts
JstnMcBrd: Ah, good question. I wanted to make the wrapper function take the same parameters as `Console.debug`, which are `(message?: any, ...optionalParams: any[])`
image

But TS didn't like me using the `any` type.
image
image

One of the "Quick fix" suggestions was using the `unknown` type (which requires type-checking on every usage - not fun) or the `never` type (which fixed all the problems). The `never` type description says "this is useful when instantiating generic-type parameters that you don't need to know the type of", which is perfect for `Console.debug`.

src/logger.ts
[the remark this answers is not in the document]
AverageHelper: There are a couple of ways. For one thing, unbound methods are not an issue in a test environment if the method isn't really a function, but a `jest.Mock` instance, and if TypeScript properly knows that. You _could_ export a named `logger` with the properties of `Console`, or at least a basic interface. The version in `__mocks__` would do the same, except its methods would all be `jest.fn()`. Then in the test, calling `jest.mock('./logger')` and then importing the logger would return that mocked instance. (You know this.) To reference the methods, you'd do something like this:
```ts
// *.test.ts
jest.mock('./logger');
import { logger } from './logger';
const mockLoggerError = logger.error as jest.Mock;

// ...
expect(mockLoggerError).toHaveBeenCalledWith(...);
```
This doesn't \*really\* reference an unbound method, since TypeScript knows `jest.Mock` will never use `this`, and this holds true at runtime since you've mocked the module. The extra typecast just tells TypeScript what's going to happen at runtime in this test. This way, you never need to export the method functions in either `logger.ts` or its mock, just that named `logger` object.

There are other ways to skin this cat as well:
- Export `logger` on the same level as its methods, import the methods directly in tests
- Remove the default export and use `import * as logger` to use the logger. This would glob together the methods defined in `logger.ts` (which are exported by name at the moment) into a new `logger` interface, a subset of `Console`, and one that we always know is mocked. We don't need to worry about adding mocks for other `Console` APIs, because there would be no `Console` APIs exported! (This also leaves room for naming the `logger` interface something else, but we can fix that in other ways. At least finding such cases would be as easy as searching for "import * as` in the codebase.)
- Probably other ways that I'm too tired to think up right now

src/logger.ts
AverageHelper: I'm not sure `unknown` is a bad idea here. The only thing that would need to do type-checking is the function's implementation, but since the value goes right to `console.debug`, which takes a parameter of type `any`, no type checking is necessary. If anything, `never` prevents _anything_ from being passed in without casting the function to something else first

JstnMcBrd: Ah, you're right, I didn't try to use the wrapper functions in a normal script until now. `unknown` works much better.

src/logger.ts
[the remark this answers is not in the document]
JstnMcBrd: Alas, the unbound methods issue remains
image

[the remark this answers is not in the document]
JstnMcBrd: And it'd be very messy to put `// eslint-disable-next-line @typescript-eslint/unbound-method` in every test script that uses the logger...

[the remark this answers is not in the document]
JstnMcBrd: Your last two suggestions are actually mutually inclusive - we can do both at the same time. That'd also allow us to define our own logger interface with any number of endpoints (if we so wished) and allow `logger.ts` to take care of redirecting the output.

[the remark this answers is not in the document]
AverageHelper: I must have been mistaken about that unbound-method issue. I wonder if we can configure `eslint-plugin-jest` to override the `@typescript-eslint` error when it makes sense, something like [this](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/unbound-method.md), perhaps. Might even consider disabling `@typescript-eslint/unbound-method` entirely for test files.

You're right tho, if we export our own functions and import with `import * as`, then there'd be no unbound method problem.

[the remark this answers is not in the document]
JstnMcBrd: I tried that jest eslint rule, and it didn't work.
`'@typescript-eslint/unbound-method': 'off'` disabled the error, but
`'jest/unbound-method': 'error'` just threw the error instead.
Apparently it couldn't accurately identify the casting as safe.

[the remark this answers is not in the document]
JstnMcBrd: And yeah, I disabled `@typescript-eslint/unbound-method` for test files at first, and that definitely worked.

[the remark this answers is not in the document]
AverageHelper: Fine by me. I don't anticipate we'll be doing much with actual unbound methods, just these cases with test files where the method is mocked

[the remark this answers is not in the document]
JstnMcBrd: I think I'll try the `import *` method. I like the idea of separating our implementation of `logger` from the interface of `Console`, so in the future, we could change the logger to perform any behavior we wish, independent of the Console endpoints.

(Plus that'll reduce the number of lines needed for exporting and importing, and I wanna improve my +/- ratio)

AverageHelper (dismissed):

ZYancey (approved):

AverageHelper (approved):

SpencerHastings (approved):
🐧

PR #41, opened 2022-10-04
Made `EventHandler` generic over its `name` property

AverageHelper:
Based on a discussion in https://github.com/BYU-CS-Discord/CSBot/pull/38#discussion_r985288404

The goal here is to remove the need for us to look up the params of each event when writing new event handlers, or maintaining existing ones. This makes the syntax of registering handlers slightly more complicated, but since that code is infrequently modified, I think a few words of boilerplate aren't too bad.

### DIFF
--- a/src/events/index.test.ts
+++ b/src/events/index.test.ts
@@ -35,11 +35,12 @@ import { _add, allEventHandlers, registerEventHandlers } from './index';
 
 describe('allEvents', () => {
 	test('index is not empty', () => {
-		expect(allEventHandlers.size).toBeGreaterThan(0);
+		expect(Object.keys(allEventHandlers).length).toBeGreaterThan(0);
 	});
 
 	test('fails to install another event handler with the same name', () => {
-		expect(() => _add({ name: 'error' } as unknown as EventHandler)).toThrow(TypeError);
+		const mockErrorHandler = { name: 'error' } as unknown as EventHandler<keyof ClientEvents>;
+		expect(() => _add(mockErrorHandler)).toThrow(TypeError);
 	});
 
 	test('properly registers events', () => {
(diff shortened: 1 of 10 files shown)


src/helpers/onEvent.test.ts
```diff
@@ -0,0 +1,20 @@
+import { onEvent } from './onEvent';
+
+describe('Creating event handlers', () => {
+	// We shouldn't have to worry about testing with poorly-formatted arguments.
+	// Since this is an internal tool, not one for a public SDK, we expect
+	// TypeScript to yell at any dev silly enough to include too many or too
+	// few arguments, or arguments of the wrong type.
+
+	test('creates a proper event handler', () => {
+		const execute = (): void => undefined;
```
JstnMcBrd, line 10: I'm pretty sure the event handler for `ready` receives a `client` parameter. Did TypeScript not yell at you about that?

src/events/index.test.ts
```diff
@@ -51,20 +52,20 @@ describe('allEvents', () => {
 		// Be sure to clear all the auto-added event handlers first, or else they'll mess up our count.
 		// Casting a read-only list into a regular list is bad practice, but this is for testing purposes.
 		// Don't do this at home.
-		(allEventHandlers as Map<keyof ClientEvents, EventHandler>).clear();
+		(allEventHandlers as Map<string, unknown>).clear();
 
-		const fakeReadyEvent: EventHandler = {
+		const fakeReadyEvent: EventHandler<'ready'> = {
 			name: 'ready',
```
JstnMcBrd, line 58: Is this line unnecessary when EventHandler defines `name = T`? (Same for `messageCreate` below)

JstnMcBrd (changes requested):
Is there any way to make EventHandler typed by `keyof ClientEvents` by default? Having to put `EventHandler<keyof ClientEvents>` everywhere you use `EventHandler` feels messy.

src/events/index.test.ts
AverageHelper: You'd think so, but TypeScript throws fits when I try:

Screen Shot 2022-10-05 at 15 32 54

AverageHelper: That's why the helper function, to reduce duplicating the `name` value

AverageHelper: Making `keyof ClientEvents` the default fixes this tho, since we cast these object to `EventHandler<keyof ClientEvents>` in the next statement anyway.

src/helpers/onEvent.test.ts
AverageHelper: You are correct that this function would receive a `client` parameter, but because that parameter is never used, and it is the only one, it may be omitted here. This is valid TypeScript.

This lambda technically _would_ receive an argument of type `Client<true>`, which may be found in its `arguments` list at runtime if we were to check that, but listing that param here is optional.

src/events/index.test.ts
```diff
@@ -51,7 +49,7 @@ describe('allEvents', () => {
 		// Be sure to clear all the auto-added event handlers first, or else they'll mess up our count.
 		// Casting a read-only list into a regular list is bad practice, but this is for testing purposes.
 		// Don't do this at home.
-		(allEventHandlers as Map<keyof ClientEvents, EventHandler>).clear();
+		(allEventHandlers as Map<string, unknown>).clear();
```
JstnMcBrd, line 52: Why unknown?

AverageHelper: Since we only wanna clear the Map, we don't care much about what's in it. This method saves us a type import

JstnMcBrd (approved):

ZYancey (approved):
My boi AvgHelper Carrying this whole team

PR #44, opened 2022-10-08
Use Builders for Commands

SpencerHastings:
This should make commands be fully functional (subcommands, context menu commands, etc) and easier to implement. The unit tests should also be passing.

### DIFF
--- a/src/commands/xkcd.ts
+++ b/src/commands/xkcd.ts
@@ -74,7 +74,7 @@ const builder = new SlashCommandBuilder()
 
 export const xkcd: GlobalCommand = {
 	requiresGuild: false,
-	commandBuilder: builder,
+	info: builder,
 
 	// entry point for command execution
 	async execute({ options, reply, sendTyping }) {
--- a/src/commands/xkcd.ts
+++ b/src/commands/xkcd.ts
@@ -65,7 +65,6 @@ async function _getComic(endpoint: string | number): Promise<GetComicResponse> {
 const builder = new SlashCommandBuilder()
 	.setName('xkcd')
 	.setDescription('Fetches the most recent xkcd comic, or a selected one.')
-	.setDMPermission(true)
 	.addIntegerOption(option =>
 		option
 			.setName('number')
--- a/src/commands/xkcd.ts
+++ b/src/commands/xkcd.ts
@@ -1,4 +1,4 @@
-import { ApplicationCommandOptionType, EmbedBuilder } from 'discord.js';
+import { EmbedBuilder, SlashCommandBuilder } from 'discord.js';
 import axios from 'axios';
 
 import { getLogger } from '../logger';
@@ -61,17 +61,20 @@ async function _getComic(endpoint: string | number): Promise<GetComicResponse> {
 	}
 }
 
+const builder = new SlashCommandBuilder()
+	.setName('xkcd')
+	.setDescription('Fetches the most recent xkcd comic, or a selected one.')
+	.setDMPermission(true)
+	.addIntegerOption(option =>
+		option
+			.setName('number')
+			.setDescription('The index number of the comic you would like to have')
+			.setMinValue(1)
+	);
+
 export const xkcd: GlobalCommand = {
-	name: 'xkcd',
-	description: 'Fetches the most recent xkcd comic, or a selected one.',
 	requiresGuild: false,
-	options: [
-		{
-			name: 'number',
-			description: 'The index number of the comic you would like to have',
-			type: ApplicationCommandOptionType.Integer,
-		},
-	],
+	commandBuilder: builder,
 
 	// entry point for command execution
 	async execute({ options, reply, sendTyping }) {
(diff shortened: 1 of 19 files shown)


SpencerHastings:
Also if ppl prefer it, the command builder can be easily inlined in the command files.

JstnMcBrd (changes requested):
Add your name to the list of developers in the README.

JstnMcBrd:
Also, is the `.idea` folder necessary? Seems like more clutter.

SpencerHastings:
> Also, is the `.idea` folder necessary? Seems like more clutter.

Is pretty much the same as the .vscode

src/commands/xkcd.ts
```diff
@@ -61,17 +61,20 @@ async function _getComic(endpoint: string | number): Promise<GetComicResponse> {
 	}
 }
 
+const builder = new SlashCommandBuilder()
+	.setName('xkcd')
+	.setDescription('Fetches the most recent xkcd comic, or a selected one.')
+	.setDMPermission(true)
```
AverageHelper, line 67: I wonder what would happen if `setDMPermission` receives `true`, but `requiresGuild` is also set to `true`. Before, TypeScript would prevent that from happening, and I'd even been planning to refactor to make `requiresGuild` obsolete by using the `dmPermission` field instead, but with that value now in the builder, I'm not sure how to read it statically and send the right context without keeping the `requiresGuild` field.

SpencerHastings: Hmm, I did check and see that the DM permission is true by default for global commands, so I'm probably going to remove that. I'll see what happens if a guild command has the dm permission set to true and if there is a good static way to catch that. I think I have an idea.

SpencerHastings: @AverageHelper Is your idea to be that commands with the dm permission are global and ones without are guild commands? I think it might be better to keep the types separate and find a way to just deny guild commands that try to set the dm permission to true.

Also I tested it, if you do `setDMPermission(true)` on a guild command, it doesn't complain but you can't use it in a DM. So imo it would be a non-issue, but I can see if I can find a way to make a unit test for it so that such commands can't be merged in regardless.

AverageHelper: The idea is that most commands that require guild information will never have a good fallback for when they don't have that information (fetching guild member stuff, channels, etc.) so I intended for the two worlds to receive a different Context type. Commands that require guild stuff should only be accessible there, so the context can have `guild`, `member`, etc. populated with something non-null always.

Ideally, there'd be a way to tell TypeScript and Discord that certain commands require guild info, and have that context statically known so that command implementers don't have to think about handling edge cases that'll never be.

SpencerHastings: My solution though would probably be to keep the `requiresGuild `boolean around, but it really just determines if a command is a guild command or a global command. Maybe it should be renamed to `isGuildCommand`. Then the dmPermission can be set on global commands still.

My worry for using the dmPermission field for determining if a command is a global or guild command is it prevents global commands from being not allowed in DMs. Also it will be nice to control what is a guild command and what is a global command, cause of the limits imposed on the amount of both.

SpencerHastings: Also I was able to write a unit test that prevents a command with requiresGuild: true from having the dm permission set to true, if that seems like a good part of the solution.

SpencerHastings:
> Add your name to the list of developers in the README.

Is done (not sure why discord doesn't allow replying to comments that aren't attached to code)

src/commands/xkcd.ts
AverageHelper: That would be helpful.

What if we renamed `requiresGuild` to `dmPermission`, inverted its value in the existing command definitions (flip `true` to `false` where its used, and vice versa), and used that value to call `setDMPermission` on the builder when we deploy?

AverageHelper: That way there'd be less code duplication. Could still use `setDMPermission` in the builder, but it wouldn't work since the one on the `GuildCommand` or `GlobalCommand` object would take precedence

JstnMcBrd (dismissed):

AverageHelper (dismissed):

src/commands/xkcd.ts
AverageHelper: I'm okay pushing this down the road. In the meantime, we can manually check in reviews that command authors aren't accidentally setting `setDMPermission` to the same value as `requiresGuild`.

AverageHelper (dismissed):

ZYancey (approved):

JstnMcBrd (changes requested):
Renaming `commandBuilder` to be more intuitive (making a note for later)

JstnMcBrd (approved):
Resolving my change request above.
No need to include me as a contributor on this merge haha

AverageHelper (approved):
Too late, yer a contributor now :P

JstnMcBrd:
You can delete the co-contributor note in the merge commit, and GitHub won't list them. Too late now though.

AverageHelper:
You... _did_ contribute tho

PR #46, opened 2022-10-09
Added a 'Fix Twitter Links' context menu item

### LINKED ISSUE #45: "vxtwitter" command
A message-context-menu command to convert twitter.com links to vxtwitter.com links in an ephemeral response.

AverageHelper:
- New context menu item sends an ephemeral message with better versions of Twitter links
- Adds basic bones for context-menu commands, similar to the way chat-input (slash) commands work
- Fixed some cases of hard-coded formatting, using [Discord.js's tools](https://discordjs.guide/popular-topics/builders.html#links)

Closes #45

### DIFF
--- a/src/helpers/actions/deployCommands.ts
+++ b/src/helpers/actions/deployCommands.ts
@@ -22,11 +22,18 @@ export async function deployCommands(client: Client<true>): Promise<void> {
 	logger.info(`Syncing ${commands.length} command(s)...`);
 
 	const guildCommands: Array<GuildedCommand> = [];
-	const globalCommands: Array<GlobalCommand> = [];
+	const globalCommands: Array<GlobalCommand | ContextMenuCommand> = [];
 	for (const cmd of commands) {
-		if (cmd.requiresGuild) {
+		if (
+			(cmd.type === undefined || cmd.type === ApplicationCommandType.ChatInput) &&
+			cmd.requiresGuild
+		) {
 			guildCommands.push(cmd);
-		} else {
+		} else if (
+			cmd.type === ApplicationCommandType.Message ||
+			cmd.type === ApplicationCommandType.User ||
+			!cmd.requiresGuild
+		) {
 			globalCommands.push(cmd);
 		}
 	}
@@ -44,12 +51,12 @@ export async function deployCommands(client: Client<true>): Promise<void> {
 }
 
 async function prepareGlobalCommands(
-	globalCommands: NonEmptyArray<GlobalCommand>,
+	globalCommands: NonEmptyArray<GlobalCommand | ContextMenuCommand>,
 	client: Client<true>
 ): Promise<void> {
 	logger.info(
 		`${globalCommands.length} command(s) will be set globally: ${JSON.stringify(
-			globalCommands.map(cmd => `/${cmd.name}`)
+			globalCommands.map(cmd => `${cmd.name}`)
 		)}`
 	);
 	logger.debug(`Deploying all ${globalCommands.length} global command(s)...`);
@@ -67,7 +74,7 @@ async function prepareGuildedCommands(
 ): Promise<void> {
 	logger.info(
 		`${guildCommands.length} command(s) require a guild: ${JSON.stringify(
-			guildCommands.map(cmd => `/${cmd.name}`)
+			guildCommands.map(cmd => `${cmd.name}`)
 		)}`
 	);
 	const oAuthGuilds = await client.guilds.fetch();
@@ -94,7 +101,7 @@ async function prepareCommandsForGuild(
 	}
 }
 
-function discordCommandPayloadFromCommand(cmd: Command): ApplicationCommandDataResolvable {
+function discordCommandPayloadFromCommand(cmd: ChatInputCommand): ApplicationCommandDataResolvable {
 	const payload: ApplicationCommandData = {
 		description: cmd.description,
 		type: cmd.type ?? ApplicationCommandType.ChatInput,
--- a/src/helpers/positionsOfUriInText.ts
+++ b/src/helpers/positionsOfUriInText.ts
@@ -1,0 +1,30 @@
+export interface Range {
+	start: number;
+	end: number;
+}
+
+/**
+ * Returns an array of index positions in the given string that may
+ * encapsulate URLs, or `null` if no URLs were found.
+ */
+export function positionsOfUriInText(str: string): NonEmptyArray<Range> | null {
+	const uri =
+		/https?:\/\/(www\.)?[-a-zA-Z0-9@:%._+~#=]{1,256}\.[a-zA-Z0-9()]{1,6}\b([-a-zA-Z0-9()@:%_+.~#?&//=]*)/gu;
+
+	let results: NonEmptyArray<Range> | null = null;
+	let match: RegExpExecArray | null = null;
+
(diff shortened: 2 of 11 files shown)


tsconfig.base.json
```diff
@@ -19,7 +19,7 @@
 		"noUncheckedIndexedAccess": true,
 		"typeRoots": ["./node_modules/@types", "./src"],
 		"types": ["node"],
-		"skipLibCheck": true,
+		"skipLibCheck": true, // disable this once our deps get their act together
```
JstnMcBrd, line 22: What's this for?

JstnMcBrd (approved):

tsconfig.base.json
AverageHelper: `skipLibCheck` disables type checking on _every_ .d.ts file. With this disabled, our own .d.ts files get type checking, which is nice when I want static analysis on my `Command` and `CommandContext` family of interfaces. While developing these changes, there've been times when I messed up and accidentally ended up with implicit `any` types that ESLint couldn't catch, because I'd misspelled something in a .d.ts file. Disabling this caught those, and that's great.

But discord.js and undici, probably a few others, are deps or sub-deps of ours whose .d.ts files do _not_ pass type checks. I wish I knew a good way to tell TS to skip checking libs in node_modules, but the prevailing wisdom at the moment is that [the TypeScript ecosystem is young](https://stackoverflow.com/a/56268670), and hasn't figured out their crap yet, so meantime we should enable `skipLibCheck` if our deps are silly.

The note here is for me to periodically check if our deps have worked out their interfaces, maybe to make a PR on them once I get mad enough, and occasionally check to see if TS has added an option to disable node_modules typechecks.

AverageHelper:
I'm not totally happy with the interop between context menu commands and their builders. I have duplicated the `type` property from the builder into the command,  for the sake of separating the execution contexts. TypeScript has no static info on whether a `ContextMenuCommandBuilder` is for a message or a user.

AverageHelper:
I have an idea.

SpencerHastings:
This would be a second command, but making a `/twit` command that automatically converts to the vxtwitter link would be neat. 

An alternative also, if someone posts a twitter link without <> around it to suppress the embed, the bot automatically deletes the embed from their post and then posts a followup with the vxtwitter link.

AverageHelper:
Great ideas! Probably good for future PRs.

As for users running this on their own message... why they wouldn't just edit their own message and add in the two characters to fix their link is beyond me XD

src/helpers/actions/deployCommands.ts
```diff
@@ -16,12 +17,17 @@ export async function deployCommands(client: Client<true>): Promise<void> {
...
 	const guildCommands: Array<GuildedCommand> = [];
-	const globalCommands: Array<GlobalCommand> = [];
+	const globalCommands: Array<GlobalCommand | ContextMenuCommand> = [];
 	for (const cmd of commands) {
-		if (cmd.requiresGuild) {
-			guildCommands.push(cmd);
-		} else {
+		if (
+			('type' in cmd &&
+				(cmd.type === ApplicationCommandType.Message ||
```
Plyb, line 24: Why are context menu commands(?) always global?

src/helpers/positionsOfUriInText.ts
```diff
@@ -0,0 +1,30 @@
...
+ * encapsulate URLs, or `null` if no URLs were found.
+ */
+export function positionsOfUriInText(str: string): NonEmptyArray<Range> | null {
+	const uri =
+		/https?:\/\/(www\.)?[-a-zA-Z0-9@:%._+~#=]{1,256}\.[a-zA-Z0-9()]{1,6}\b([-a-zA-Z0-9()@:%_+.~#?&//=]*)/gu;
+
+	let results: NonEmptyArray<Range> | null = null;
+	let match: RegExpExecArray | null = null;
+
+	while ((match = uri.exec(str))) {
```
Plyb, line 17: nit: extra parens? Or does this do something I'm ont aware of.

Plyb (approved):

src/helpers/actions/deployCommands.ts
AverageHelper: Because I can't be bothered to make guild-only ones yet 😅. I literally only need this type tree for this one command, and this command can run anywhere. We might not even do more context menu commands, let alone ones that commonly need guild guarantees, certainly not when our main focus is parity with Ze Kaiser.

src/events/interactionCreate.test.ts
```diff
@@ -10,15 +24,40 @@ jest.mock('../commands', () => ({
...
+	requiresGuild: false,
+	execute: mockGlobalExecute,
+};
+mockAllCommands.set(mockUserContextMenuCommand.info.name, mockUserContextMenuCommand);
+
 const mockGuildedExecute = jest.fn();
-const mockGuildedCommand: Command = {
-	info: new SlashCommandBuilder().setName('guilded-test').setDescription('lolcat'),
+const mockGuildedCommand: ChatInputCommand = {
+	info: new SlashCommandBuilder() //
```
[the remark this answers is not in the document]
AverageHelper: Keep reading. That `//` tells Prettier not to squish this method chain onto one line

src/events/interactionCreate.test.ts
```diff
@@ -10,15 +24,40 @@ jest.mock('../commands', () => ({
 
 // Create two mock commands to track handler behavior
 const mockGlobalExecute = jest.fn();
-const mockGlobalCommand: Command = {
-	info: new SlashCommandBuilder().setName('global-test').setDescription('lolcat'),
+const mockGlobalCommand: ChatInputCommand = {
+	info: new SlashCommandBuilder() //
```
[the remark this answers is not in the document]
AverageHelper: That // is the cleanest way I've found to tell Prettier not to squish this method chain onto one line. It has the added side-effect of angling toward the next line, if your eyes use it like a slide ^^

src/helpers/positionsOfUriInText.ts
AverageHelper: ESLint thinks the assignment expression is accidental unless I tell it its intentional by wrapping it in parens. The goal of this line is to run the loop if a match is found. Could probs be clearer, but I didn't wanna duplicate the assignment expression lol

Thoughts on a clearer syntax?

src/events/interactionCreate.test.ts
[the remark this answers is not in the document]
AverageHelper: I realize now you might have meant "empty comment." Sorry for my rude "keep reading" comment, lol

src/events/interactionCreate.test.ts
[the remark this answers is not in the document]
AverageHelper: I'm open to suggestions on how to improve this tho

JstnMcBrd (dismissed):

JstnMcBrd:
Oh, we should probably increment the version number with this update.

AverageHelper:
uright

JstnMcBrd (approved):

src/helpers/positionsOfUriInText.ts
Plyb: ah, that makes sense now

Plyb (approved):

PR #50, opened 2022-10-14
First go at reaction duplication

### LINKED ISSUE #24: Reaction duplication
When users add reactions to messages, the bot will some fraction of the time, also react with that same reaction.

`/set-reaction-rate <rate> [emoji]` sets how often the bot reacts. If `emoji` is not specified, this will set the default reaction rate for all reactions, which can be overriden on a reaction-by-reaction basis by specifying `emoji`. `rate` is a simple fractional number like `0.1`. If this command has not been run yet, set the default reaction rate to `0.05`.

`/get-reaction-rate [emoji]` prints the set rate for either the default or a specific emoji. If no specific rate is set for an emoji, show the default and indicate that it is is the default.

AverageHelper:
New stuff:
- The bot now duplicates some initial message reactions.
  - :star: is ignored
  - 1 in 5 chance for :same: and :no_u:
  - 1 in 100 for every other emoji
  - Emoji names are case-normalized (e.g. any capitalization of :same: will do)

I anticipate we can make these odds configurable once we have the database fleshed out. We'd need at least:
- Some place to store the default reaction chance
- Some place to store the names and off-chances of each special emoji
- Some way to tie those emoji to the eventual X-board (to prevent the bot from affecting those stats) without excess database I/O

Closes #24

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,9 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## 0.4.0 - 2022-10-11
+## 0.5.0 - 2022-10-13
 ### Added
-- A message context menu command to fix Twitter links using vxtwitter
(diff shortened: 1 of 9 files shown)


JstnMcBrd:
Might be best to put the `DEFAULT_CHANCE` and `odds` in the constants file, at least for now.

JstnMcBrd (approved):

Plyb (approved):

PR #54, opened 2022-10-17
Buttons

Plyb:
Adds button interaction handling to the bot. No actual buttons are implemented yet.
 - Separates out the code that can be shared between command and button handling
 - button handlers create their own ButtonBuilders allowing the styling and customId of the button to be consistent between usages
 - `Button`s can be consumed by `Command`s.

### DIFF
--- a/src/@types/InteractionContext.d.ts
+++ b/src/@types/InteractionContext.d.ts
@@ -1,4 +1,5 @@
 import type {
+	ButtonInteraction,
 	ChatInputCommandInteraction,
 	Client,
 	CommandInteraction,
@@ -12,54 +13,43 @@ import type {
 	Message,
 	MessageContextMenuCommandInteraction,
 	MessageReplyOptions,
+	RepliableInteraction,
 	Snowflake,
 	User,
 	UserContextMenuCommandInteraction,
 } from 'discord.js';
 
 declare global {
-	interface BaseCommandContext {
+	// We need to use a type intersection because BaseInteraction doesn't have reply methods
+	type RepliableInteraction = ButtonInteraction | CommandInteraction;
+
+	interface InteractionContext {
 		/** Where the command was invoked. */
 		readonly source: 'guild' | 'dm';
 
-		/** The command invocation interaction. */
-		readonly interaction: CommandInteraction;
+		/** The triggering interaction. */
+		readonly interaction: RepliableInteraction;
 
 		/** Our own signed-in Discord client. */
 		readonly client: Client<true>;
 
-		/** The guild in which the command was invoked. */
+		/** The guild in which the interaction was invoked. */
 		readonly guild: Guild | null;
 
-		/** The ID of the channel in which the command was invoked. */
-		readonly channelId: Snowflake;
+		/** The ID of the channel in which the interaction was invoked. */
+		readonly channelId: Snowflake | null;
 
-		/** The channel in which the command was invoked. */
+		/** The channel in which the interaction was invoked. */
 		readonly channel: GuildTextBasedChannel | DMChannel | null;
 
-		/** The user who invoked the command. */
+		/** The user who invoked the interaction. */
 		readonly user: User;
 
-		/** The guild member who invoked the command. */
+		/** The guild member who invoked the interaction. */
 		readonly member: GuildMember | null;
 
-		/** The ID of the interaction target. Only available for context menu commands. */
-		readonly targetId: Snowflake | null;
-
-		/** The user that the interaction targets. Only available for context menu commands. */
-		readonly targetUser: User | null;
-
-		/** The guild member that the interaction targets. Only available for context menu commands. */
-		readonly targetMember: GuildMember | null;
-
-		/** The message that the interaction targets. Only available for context menu commands. */
-		readonly targetMessage: Message | null;
-
-		/** The UNIX time at which the command was invoked. */
+		/** The UNIX time at which the interaction was invoked. */
 		readonly createdTimestamp: number;
-
-		/** The options that were given to the command. Not available for context menu commands. */
-		readonly options: ReadonlyArray<CommandInteractionOption<'cached'>> | null;
 
 		/** Instructs Discord to keep interaction handles open long enough for long-running tasks to complete. */
 		prepareForLongRunningTasks: (ephemeral?: boolean) => void | Promise<void>;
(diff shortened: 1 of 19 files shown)


src/@types/InteractionContext.d.ts
```diff
@@ -17,49 +17,34 @@ import type {
...
-		readonly targetId: Snowflake | null;
-
-		/** The user that the interaction targets. Only available for context menu commands. */
-		readonly targetUser: User | null;
-
-		/** The guild member that the interaction targets. Only available for context menu commands. */
-		readonly targetMember: GuildMember | null;
-
-		/** The message that the interaction targets. Only available for context menu commands. */
-		readonly targetMessage: Message | null;
```
JstnMcBrd, line 55: Did you mean to move these properties `BaseContextMenuCommandContext`?

src/@types/InteractionContext.d.ts
```diff
@@ -17,49 +17,34 @@ import type {
...
 declare global {
-	interface BaseCommandContext {
+	interface InteractionContext {
 		/** Where the command was invoked. */
 		readonly source: 'guild' | 'dm';
 
-		/** The command invocation interaction. */
-		readonly interaction: CommandInteraction;
+		/** The triggering interaction. */
+		readonly interaction: RepliableInteraction;
```
JstnMcBrd, line 25: I think you may have forgotten to update your imports when you switched these out.
I still see `CommandInteraction` and I don't see `RepliableInteraction`.

src/events/interactionCreate.ts
```diff
@@ -165,3 +145,63 @@ async function handleInteraction(interaction: CommandInteraction): Promise<void>
...
+	const buttonContext: ButtonContext = {
+		...context,
+		component: interaction.component,
+		message: interaction.message,
+	};
+
+	return await button.execute(buttonContext);
+}
+
+async function handleInteraction(interaction: RepliableInteraction): Promise<InteractionContext> {
```
JstnMcBrd, line 172: Maybe this should be named something like `generateContext`?

src/events/interactionCreate.ts
```diff
@@ -20,8 +28,18 @@ export const interactionCreate = onEvent('interactionCreate', {
...
+			if (interaction.user.id === interaction.client.user.id) return;
+
 			if (interaction.isCommand()) {
-				await handleInteraction(interaction);
+				const context = await handleInteraction(interaction);
+				await handleCommandInteraction(context, interaction);
+			}
+
+			if (interaction.isButton()) {
+				const context = await handleInteraction(interaction);
```
JstnMcBrd, line 41: These 2 context-generation lines can probably be moved out of the if-statements so they aren't duplicated

src/events/interactionCreate.ts
```diff
@@ -165,3 +145,63 @@ async function handleInteraction(interaction: CommandInteraction): Promise<void>
...
+	context: InteractionContext,
+	interaction: ButtonInteraction
+): Promise<void> {
+	logger.debug(`User ${logUser(interaction.user)} pressed button: '${interaction.customId}'`);
+
+	const button = allButtons.get(interaction.customId);
+	if (!button) {
+		logger.warn(`Received request to execute unknown button with id '${interaction.customId}'`);
+		return;
+	}
```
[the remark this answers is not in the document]
Plyb: This was based on a similar pattern in `handleCommandInteraction`. It might be good to just flag this for now. Avg has made a change to this in #63 that would be a good pattern to follow, but that needs to be merged first, or if this gets merged first, the error handling can be added on his branch. I'll put a TODO on it.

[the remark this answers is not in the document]
AverageHelper: #63 only needs one more review

ZYancey (dismissed):

src/events/interactionCreate.ts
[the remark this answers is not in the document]
AverageHelper: Merged.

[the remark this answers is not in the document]
JstnMcBrd: Conflict resolution time 😭

src/@types/InteractionContext.d.ts
```diff
@@ -1,4 +1,5 @@
 import type {
+	ButtonInteraction,
 	ChatInputCommandInteraction,
```
AverageHelper, line 3: ESLint tells me that `ButtonInteraction` is never used.

```suggestion
import type {
	ChatInputCommandInteraction,
```

src/@types/InteractionContext.d.ts
```diff
@@ -213,6 +223,12 @@ declare global {
 		readonly targetMessage: Message;
 	}
 
+	/** Information relevant to button presses */
+	interface ButtonContext extends InteractionContext {
+		component: APIButtonComponent;
```
AverageHelper, line 228: Please `APIButtonComponent`.

src/@types/InteractionContext.d.ts
```diff
@@ -112,6 +99,29 @@ declare global {
 		) => Promise<Message | boolean>;
 	}
 
+	interface BaseCommandContext extends InteractionContext {
+		/** The command invocation interaction. */
+		readonly interaction: CommandInteraction;
```
AverageHelper, line 104: With `skipLibCheck` disabled, TypeScript complains about this field not matching the one defined in the parent interface. Not sure the best way to fix this, since I don't know this new button abstraction well at all. Might need to decide whether our commands need a `CommandInteraction` or a `RepliableInteraction` in order to work properly, and use that for both. That or somehow decouple `BaseCommandContext` from `InteractionContext`, perhaps by creating a new parent context interface, or simplifying `BaseCommandContext` and raising that to be the parent interface instead.

It seems we only use `BaseCommandContext` internally in this file (except in the `talk` command, which could use a more specific interface like `RepliableInteraction` instead.

AverageHelper (changes requested):

src/events/interactionCreate.ts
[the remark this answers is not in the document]
AverageHelper: :excellent:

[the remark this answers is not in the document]
Plyb: Luckily this ended up being a lot less hairy than I expected heh

src/@types/InteractionContext.d.ts
Plyb: This ended up being pretty simple to resolve by just using `BaseInteraction` as the type for interaction on `InteractionContext`. Typescript is smart enough to figure the rest out.

src/@types/InteractionContext.d.ts
Plyb: Thanks, I'll be careful to disable `skipLibCheck` to check this kind of thing in the future.

src/@types/InteractionContext.d.ts
Plyb: fixed

AverageHelper (changes requested):

AverageHelper (approved):

ZYancey (approved):

PR #57, opened 2022-11-11
Dectalk

### LINKED ISSUE #48: Dectalk-powered TTS command
Dectalk code: https://github.com/Babakinha/Dectalk
Dectalk example: https://www.youtube.com/watch?v=Hv6RbEOlqRo

A /speak or /talk or /say or /dectalk command
If used in a text channel, sends a .wav file as an attachment (which Discord will automatically embed)
If used in a voice channel chat, speaks in the voice channel

JstnMcBrd:
(Closes #48 )
(Relies on #58 )

This one has been a long time coming.

### Dectalk changes:
- added a /talk slash command 
  - if used in a text channel, it sends a .wav file attachment
  - if used in a voice channel chat, it joins the voice channel, speaks the message, and then leaves
  - also has an option to change the voice of the speaker - Paul, Kid, Ursula, etc
- added a Talk context command so users can right-click on a message and have the bot say the message
  - makes it easier to repeat the same command too, because the bot will always print out the message it speaks, so you can just right-click on the bot's last message to repeat it
  - context command reuses the same code as the slash command
- tests for both slash command & context command
- added a new Linux dependency in the docker image

### Other changes:
- added new command error reporting system, with some privacy protections
- added unit tests for the new command error reporting system
- modified all other commands to use the new error reporting system (made them a lot simpler)
- updated tests for all other commands to reflect the new error reporting system (made them a lot simpler too)
- fixed issue where only 1 docker container could run at a time
- updated version and changelog


src/commands/talk.test.ts
```diff
@@ -0,0 +1,116 @@
...
+					},
+					getString(name: string, required: boolean) {
+						name;
+						required;
+						return message;
+					},
+					getInteger(name: number) {
+						name;
+						return undefined;
+					},
```
AverageHelper, line 40: When no parameter is used, and the function's type is defined elsewhere, we don't need to list the parameters. This means we don't need useless references either:

```suggestion
					getString: () => message,
					getInteger: () => undefined,
```

src/commands/talk.ts
```diff
@@ -0,0 +1,224 @@
+// External dependencies
+import { writeFileSync, createReadStream } from 'node:fs';
+import { isNumber } from 'lodash';
```
AverageHelper, line 3: Lodash is big and expensive to load. Let's not load the whole thing ^^

```suggestion
import isNumber from 'lodash/isNumber';
```

src/commands/talk.ts
```diff
@@ -0,0 +1,224 @@
...
+	.addStringOption(option =>
+		option.setRequired(true).setName('message').setDescription('The message to speak')
+	)
+	.addIntegerOption(option => {
+		option.setRequired(false).setName('speaker').setDescription('Whose voice to use');
+
+		Object.values(Speaker)
+			.filter(isNumber)
+			.forEach(value => {
+				option = option.addChoices({ name: Speaker[value] as string, value: value });
```
AverageHelper, line 40: Better to handle the `undefined` case (using `??` or literally anything else) than to use `as` to explicitly ignore it.

src/commands/talk.ts
```diff
@@ -0,0 +1,224 @@
...
+			});
+
+		return option;
+	});
+
+export const talk: GlobalCommand = {
+	info: builder,
+	requiresGuild: false,
+	async execute(context) {
+		const options = context.interaction.options as CommandInteractionOptionResolver;
```
AverageHelper, line 50: This type cast is dangerous. The options resolver is defined to have the methods we use below if the interaction is a `ChatInputCommandInteraction`. I must have forgotten this distinction when creating the `CommandContext` types.

Meantime, checking that the interaction is a `ChatInputCommandInteraction` makes the requisite methods available, without spoopy typecasts:
```suggestion
		if (!context.interaction.isChatInputCommand()) throw new Error("...");
		const options = context.interaction.options;
```

src/commands/talk.ts
```diff
@@ -0,0 +1,224 @@
...
+		const stream = createReadStream(tempFile.name);
+
+		// create an audio resource from the temporary file
+		let resource: AudioResource;
+		try {
+			resource = createAudioResource(stream);
+		} catch (error) {
+			const error_ = error as Error;
+			// For this error, provide a more useful message
+			if (error_.message === 'FFmpeg/avconv not found!') {
```
AverageHelper, line 127: Avoid typecasts, unless there's no other way and assumptions _must_ be made.

I hope the thrown error here is an `Error` or `Error` subclass. If not, then this `instanceof` check should be replaced with something more complicated to assert that `message` exists.

```suggestion
			// For this error, provide a more useful message
			if (error instanceof Error && error.message === 'FFmpeg/avconv not found!') {
```

src/commands/talk.ts
```diff
@@ -0,0 +1,224 @@
...
+		connection.on(VoiceConnectionStatus.Disconnected, () => {
+			log('disconnected');
+			connection.destroy();
+
+			log('replying to interaction');
+			void reply({
+				content: message,
+			});
+
+			player.stop();
```
AverageHelper, line 190: Does this fire the `AudioPlayerStatus.Idle` event on the player, or should we also close the stream here?

AverageHelper (changes requested):
The reduced coverage in some places makes me sad, but I like the new just-throw-the-error paradigm. Makes handling those much easier!

Ideally, the tests would also check the content of the error thrown, similar to the way the tests used to check the reply content, rather than just "was any error thrown?". Those tests can be added in later tho

src/commands/talk.ts
AverageHelper: I'm working on a new PR to make this check unneeded. Slash commands should always get a `ChatInputCommandInteraction`.

AverageHelper: See #58

src/commands/talk.ts
JstnMcBrd: Yes, stopping the player fires the Idle event.

JstnMcBrd:
> Ideally, the tests would also check the content of the error thrown, similar to the way the tests used to check the reply content, rather than just "was any error thrown?".

I believe Jest has a way to test the message of an error using regular expressions, but I wasn't able to figure it out when I tried it. Maybe you'd have better luck than me.

src/commands/talk.ts
JstnMcBrd: I'll fix this after #58 is merged. Leaving this thread unresolved for now.

AverageHelper (approved):

ZYancey (approved):

ZYancey (approved):

AverageHelper (approved):

PR #68, opened 2023-01-04
Update command

Plyb:


src/commands/update.ts
```diff
@@ -0,0 +1,41 @@
...
+	.setDescription('Pulls the latest changes and restarts the bot')
+	.setDefaultMemberPermissions('0');
+
+export const update: GlobalCommand = {
+	info,
+	requiresGuild: false,
+	async execute({ replyPrivately, interaction }) {
+		const admin_ids = process.env['ADMINISTRATORS']?.split(',');
+		if (!admin_ids) {
+			// TODO: make this a UserMessageException
```
Plyb, line 15: UserMessageException is part of the EvilHangman PR, can't do this right now without a bunch of unnecessary hassle.

src/commands/update.ts
```diff
@@ -0,0 +1,41 @@
...
+const info = new SlashCommandBuilder()
+	.setName('update')
+	.setDescription('Pulls the latest changes and restarts the bot')
+	.setDefaultMemberPermissions('0');
+
+export const update: GlobalCommand = {
+	info,
+	requiresGuild: false,
+	async execute({ replyPrivately, user, interaction }) {
+		const admin_ids = process.env['ADMINISTRATORS']?.split(',');
```
AverageHelper, line 13: Might be worth documenting this new `ADMINISTRATORS` env var.

src/commands/update.ts
```diff
@@ -0,0 +1,41 @@
...
+
+		if (!admin_ids.includes(user.id)) {
+			// TODO: make this a UserMessageException
+			throw new Error(
+				'You do not have permission to perform this command. Contact the bot administrator.'
+			);
+		}
+
+		await replyPrivately('Updating...');
+		exec('npm run update', err => {
```
AverageHelper, line 27: What consequences might there be to this command being run several times in quick succession, by one or more users?

src/commands/update.ts
```diff
@@ -0,0 +1,41 @@
...
+				if (err) {
+					throw err;
+				}
+				await interaction.editReply('Finished updating. Restarting now.');
+				exec('npm run restart', err => {
+					if (err) {
+						throw err;
+					}
+				});
+			})();
```
AverageHelper, line 38: Lots of chained callbacks here. Could these be wrapped in `Promise`-returning functions for easier reading?

src/commands/update.ts
```diff
@@ -0,0 +1,41 @@
...
+			throw new Error(
+				'You do not have permission to perform this command. Contact the bot administrator.'
+			);
+		}
+
+		await replyPrivately('Updating...');
+		exec('npm run update', err => {
+			void (async (): Promise<void> => {
+				if (err) {
+					throw err;
```
AverageHelper, line 30: Since we invoke the function using `void`, these thrown errors are never caught. [`void`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/void) only evaluates the given expression (in this case, kicking off a `Promise`-returning anonymous function) and returns `undefined` synchronously. If an error is thrown in the process of some asynchronous task that isn't `await`-ed, that error only ever appears in the console, and cannot be handled appropriately. Might even cause a crash, depending on runtime flags.

src/commands/update.test.ts
```diff
@@ -0,0 +1,68 @@
...
+	} as unknown as User;
+	const mockReplyPrivately = jest.fn<Promise<void>, [content: string]>();
+	const mockEditReply = jest.fn<Promise<void>, [content: string]>();
+
+	let context: TextInputCommandContext;
+
+	beforeEach(() => {
+		// Overwrite the environment variable for token
+		const mockAdmins = '1,2';
+		process.env[ADMINISTRATORS_VARIABLE] = mockAdmins;
```
AverageHelper, line 33: It is customary, when overwriting globals, to store the original global and do the overwrite in `beforeEach`, then restore the global in `afterEach`. This way, tests don't need to make assumptions about shared state with respect to other tests.

AverageHelper (changes requested):

src/commands/update.ts
AverageHelper: There are several ways to handle this. In general, each invocation needs to check that it is singleton, and if it is not, abort execution. Some ideas come to mind:
- A global `boolean` variable that declares whether this command is being invoked. To avoid race conditions, that variable should be set early and checked frequently, so as to back out early in case of spoopy behavior.
- A global `number` or `bigint` variable that declares the number of running instances. Similar to the global `boolean` idea, with added information for debugging. ("Why was this thing invoked _3_ times in a row?")

We might consider storing this info at the top level of the module, letting Node disambiguate invocations within its process. A more robust model might store the info in the database, so as to be resilient to multiple Node process firing at once (slightly more likely since we now defer the runtime execution to PM2).

This is only a simple "do an update" command, but its self-surgery nature demands careful thought to risk management.

src/commands/update.ts
AverageHelper: If it were up to me, we'd even import `exec` as something scary-looking, like `__unsafeExecuteCommand`, and then wrap that in an `async` function.

src/commands/update.ts
Plyb: This is a good point. The solution I've added counts the number of invocations, and based on how JS's event loop works, we shouldn't run into any race conditions. As for cases where there are mutliple instances of the bot running, that shouldn't happen since pm2 would try to give them the same process name. If it were to happen though, nothing short of like an inter-process semaphore would work (is that a thing?), so I don't think it's worth hooking it up to the db.

As far as what the consequences are: the bot could fail to restart, possibly the `dist` code gets corrupted in some way, in which case manual intervention would be required. Considering though that this command is mainly a convenience so I (and future bot admins) don't need to manually update it every time, even if it occassionally fails that's a win. With this new check, I don't know of a way it could fail (short of the bot admin starting another bot instance outside pm2 I guess), but still

src/commands/update.ts
Plyb: obviated by the promisification

src/commands/update.ts
Plyb: done

src/commands/update.test.ts
Plyb: done

src/commands/update.ts
```diff
@@ -0,0 +1,54 @@
...
+			);
+		}
+
+		numInvocations += 1;
+		if (numInvocations > 1) {
+			throw new Error(
+				`Cannot run update, there are already ${(
+					numInvocations - 1
+				).toString()} update invocations running`
+			);
```
AverageHelper, line 33: When using a template string, calling `toString` on a `number` is redundant.

```suggestion
			throw new Error(
				`Cannot run update, there are already ${numInvocations - 1} update invocations running`
			);
```

AverageHelper (changes requested):

src/commands/update.ts
```diff
@@ -0,0 +1,52 @@
...
+			throw new Error(
+				`Cannot run update, there are already ${numInvocations - 1} update invocations running`
+			);
+		}
+
+		await replyPrivately('Updating...');
+		await execAsync('npm run update');
+		await interaction.editReply('Finished updating. Restarting now.');
+		await execAsync('npm run restart');
+		numInvocations -= 1; // In theory, we should never arrive here. Including for testing purposes.
```
AverageHelper, line 38: Shouldn't we? I figure once the commands have been `await`-ed, if they didn't throw, we should arrive here.

At any rate, this assignment should probs go in the tail end of a `try...finally` block

AverageHelper (changes requested):

src/commands/update.ts
Plyb: once `npm run restart` finished, this process will have been killed. I suppose in any case that my thought behind the comment doesn't really make sense. If we've somehow gotten here without the bot dying, we probably want to be able to run the command again. Good call on the try...finally, I'll add that.

AverageHelper: In that case, it might be prudent to wrap `execAsync('npm run restart')` in a function that returns `Promise<never>`, so that TypeScript is aware that we plan on dying before it resolves.

AverageHelper: With helpful comments explaining this weirdness, of course :P

AverageHelper: Perhaps something like this:

```ts
/**
 * Restarts the bot.
 */
async function restart(): Promise<never> {
	await execAsync('npm run restart');
	process.exit(); // so that TypeScript is *absolutely sure* we're dead after this
}
```

Plyb: I can wrap it in a function that returns `Promise<never>`, but typescript doesn't like that (according to it), the function has a reachable end point. What would you recommend doing to have it not be angry? I can have it throw after, which feels a little hacky, or maybe use an assertion comment.

AverageHelper: TIL that TypeScript doesn't realize that `await`-ing a `Promise<never>` results in unreachable code following that call. Maybe it will one day, idk. Seems a good idea nonetheless to have a `Promise` in there that semantically "`never` resolves."

AverageHelper (approved):

src/commands/update.ts
AverageHelper: That `process.exit()` should never get hit, but if it does, at least we know the bot is dead dead by the end, lol

AverageHelper (approved):

src/commands/update.ts
AverageHelper: Oh crap, the tests don't know about this. Would they even run to completion now?

Plyb: ah, good catch. No they don't

AverageHelper: I'm wondering now if we shouldn't just throw something that's like "We shouldn't have gotten here, lol"

Plyb: ok, I think I've got something that satisfies both the tests and production, and makes some sense to a reader. Throwing an error makes the tests overly complicated.

AverageHelper (approved):

gmacgre (approved):

PR #75, opened 2023-03-16
Updated version of vulnerable undici dependency

josephgeis:
npm reported a vulnerability in the `undici` dependency (of some other dependency). This updates the `package-lock.json` to use the patched version.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
 	"name": "csbot",
-	"version": "0.7.0",
+	"version": "0.8.0",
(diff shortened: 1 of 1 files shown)


AverageHelper:
I've seen this in another project of mine as well. Seems that the latest undici doesn't play well with Jest

AverageHelper:
As noted in #82, fixing this will be complicated, and may involve porting our compiled output to ESM (which may need to be done eventually anyway). I'm working on a separate PR to do this.

AverageHelper:
Will merge into main with #86

PR #76, opened 2023-03-16
Add isCASDown command

### LINKED ISSUE #74: Slash Command - Is CAS down?
`/is-cas-down`
See this [Discord thread](https://discord.com/channels/341633933894746122/1085760745595228270).

Would ping `cas.byu.edu` and respond if CAS is down.

Perhaps use [ping](https://www.npmjs.com/package/ping) node module?

josephgeis:
Implements/closes #74. I need help writing tests though.


AverageHelper:
You can add tests by following the examples in other `*.test.ts` files. You will need to mock `axios`, and consider both success and failure cases. Keep in mind that our command runner architecture knows what to do with most errors thrown, so don't worry about sending messages to users about network errors.

src/commands/isCASDown.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export const isCASDown: GlobalCommand = {
+	info: builder,
+	requiresGuild: false,
+	async execute({ reply }) {
+		const embed = new EmbedBuilder().setTitle('Is CAS Down?');
+
+		let status = statuses.down;
+
+		const req = await axios.get(statusURI).catch(() => null);
```
AverageHelper, line 26: Network errors might be useful to know. Also, our command runner architecture knows what to do with them, so best to let the throw happen.

```suggestion
		const req = await axios.get(statusURI);
```

src/commands/isCASDown.ts
```diff
@@ -0,0 +1,42 @@
...
+
+		if (req) {
+			if (req.status == 200) {
+				status = statuses.up;
+			}
+		}
```
AverageHelper, line 32:
```suggestion
		if (req?.status === 200) {
			status = statuses.up;
		}
```

AverageHelper, line 32: `req` usually means "request", whereas `res` usually means "response".

```suggestion
		const res = await axios.get(statusURI).catch(() => null);

		if (res) {
			if (res.status == 200) {
				status = statuses.up;
			}
		}
```

AverageHelper (changes requested):

src/commands/isCASDown.ts
AverageHelper: Also, a network error doesn't necessarily mean that CAS is down. We'd expect something like a successful response with a status of `500`. So catching the error and defaulting to a `null` response wouldn't be useful.

src/commands/isCASDown.ts
josephgeis: Ahh forgot about this syntax. It's been a hot second since I've worked with JS.

src/commands/isCASDown.ts
josephgeis: I made this change. Didn't realize that there was handling of it, but my thought was if there was an error, the command still needs to report it as CAS being down, so having it be `null` would trigger that. We could always revert though.

AverageHelper:
Hmm, seems that our coverage report tool doesn't like forks. Need to figure a way around that.

I wonder if the changes in #82 make a difference, or worsen the trouble...

AverageHelper:
I see a light. Will probably make a PR soon-ish to fix the CI such that forks can run tests. Once that's made and merged, this PR should begin working again!

AverageHelper:
#84 should fix the tests

AverageHelper (approved):

Plyb (dismissed):
still need tests on this one

AverageHelper (approved):

PR #82, opened 2023-03-21
Update Deps

AverageHelper:
Despite the name of this branch, we expressly _avoid_ updating discord.js for now, for the same reason that #75 currently fails CI. There's an odd issue that seems to have to do with mixing CommonJS and WASM. A lot of Undici is WASM. Not sure if that's new, but that is the case, and for some reason, Jest doesn't like WASM one bit as presently configured. Will need to port to ESM in order to fix that.

For now, let's just update some other deps.


AverageHelper:
Gonna fix those red arrows before I mark this as ready.

.github/workflows/test.yml
```diff
@@ -33,10 +33,4 @@ jobs:
         run: npm run build
 
       - name: Run unit tests
-        run: npm test
-        env:
-          CI: true
-          NODE_ENV: 'test'
-
-      - name: Report test coverage
```
JstnMcBrd, line 41: I noticed you removed `npm test`, and just replaced it with the reporting stage (`ArtiomTr/jest-coverage-report-action@v2.1.1`). Does that automatically run the tests?

AverageHelper: It does, twice. Once to get one batch of coverage, and another time on the target branch to get the other batch, to diff the coverage for the chart. If the tests fail, then it blocks the pipeline (as it should).

AverageHelper: I'd previously assumed that it gets coverage from our own test run, but no, it runs its own.

AverageHelper: You can see this in [this PR's previous CI run](https://github.com/BYU-CS-Discord/CSBot/actions/runs/4483065715/jobs/7881896098), under the "Run unit tests" action.

Plyb (approved):

.github/workflows/test.yml
AverageHelper: Actually, we've disabled this call anyway, lol

JstnMcBrd (approved):

Plyb (approved):

Issue #56, opened 2022-10-20
Slash command - retrieve emoji images


JstnMcBrd:
The purpose of the `/emoji` command would be similar to `/profile` - return the internal image used for a custom emoji.

Discord.js has this functionality with the [`Emoji.url`](https://discord.js.org/#/docs/discord.js/stable/class/Emoji?scrollTo=url) property.

AverageHelper:
Considering this is another one that can be done in devtools, and there's not an associated user to mention, should this one be ephemeral?

JstnMcBrd:
Good question. Depends on our use case.

I could see it being used for 2 different purposes: either to retrieve and download the emoji image, or to post an extra-large emoji in response to someone (similar to Ze Kaiser's tags). Image a giant `/emoji kekw` after something posts a hilarious meme, etc.

For downloading emojis, ephemeral would be best. But that would make it impossible to use it for a reaction. So I'm not sure either way. Thoughts?

AverageHelper:
@Plyb or other experienced moderators might be the best one to weigh in on the potential spam for an easy reaction-image command other than the regular tagged images

TStansel:
I have a working version of this, just need to finish tests. Is is possible to get this assigned to me?

AverageHelper:
@TStansel Feel free to fork and make a PR!

TStansel:
Just pushed a PR for this

PR #83, opened 2023-03-30
Slash command retrieve emoji images

TStansel:
The goal of this PR is the address issue #56.

In this issue there was some discussion as to if the response from the bot should be ephemeral or not. As of now, it is ephemeral but this is a simple boolean change. Because the change from ephemeral to not ephemeral is so simple, I figured that it could be discussed further in this PR or could be updated in a later PR if needed.
(@Plyb @JstnMcBrd @AverageHelper for further discussion if needed)

### DIFF
--- a/src/commands/emoji.test.ts
+++ b/src/commands/emoji.test.ts
@@ -1,0 +1,65 @@
+import { EmbedBuilder, GuildEmoji, ImageURLOptions } from 'discord.js';
+import { emoji } from './emoji';
+
+jest.mock('../logger');
+
+describe('profile', () => {
+	const mockReply = jest.fn<Promise<void>, [content: unknown]>();
+	const mockEmojiName = jest.fn<string | null, []>();
+	const mockEmojiURL = jest.fn<string | null, [options?: ImageURLOptions | undefined]>();
+
+	const testEmojiURL = 'https://example.com/emojis/1234567890/abcdef1234567890.png';
+
+	let context: TextInputCommandContext;
+	let existingEmoji;
+
+	beforeEach(() => {
+		existingEmoji = {
+			name: 'existing emoji',
+			url: testEmojiURL,
+		} as unknown as GuildEmoji;
+
+		context = {
+			interaction: {
+				options: {
+					client: {
+						emojis: {
+							cache: [existingEmoji],
+						},
+					},
+					getString: mockEmojiName,
+				},
+			},
+			reply: mockReply,
+		} as unknown as TextInputCommandContext;
+
+		mockEmojiName.mockReturnValue('existing emoji');
+		mockEmojiURL.mockReturnValue(testEmojiURL);
+	});
+
+	test('Throws an error when no input is provided', async () => {
+		mockEmojiName.mockReturnValue(null);
+		await expect(emoji.execute(context)).rejects.toThrow('You must provide an emoji to search for');
+	});
+
+	test('Throws an error when the target emoji does not exist', async () => {
+		mockEmojiName.mockReturnValue('emoji not here');
+		await expect(emoji.execute(context)).rejects.toThrow('Emoji emoji not here was not found');
+	});
+
+	test('Returns the url of the target emoji', async () => {
+		await expect(emoji.execute(context)).resolves.toBeUndefined();
+
+		expect(mockReply).toHaveBeenCalledOnce();
+		expect(mockReply).toHaveBeenCalledWith({
+			content: mockEmojiName() as string,
+			embeds: [
+				new EmbedBuilder({
+					title: mockEmojiName() as string,
+					image: { url: testEmojiURL },
+				}),
+			],
+			ephemeral: true,
+		});
+	});
+});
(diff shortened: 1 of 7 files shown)


JstnMcBrd:
Perhaps the ephemeral option could be part of the command - it would default to true, but users could choose to set it to false.

JstnMcBrd:
Also, don't forget to make maintenance changes:
- add the command to the list of commands in the README
- bump the project version in package.json
- add your changes to the CHANGELOG

TStansel:
> Perhaps the ephemeral option could be part of the command - it would default to true, but users could choose to set it to false.

I could add this, but I'm not sure if this solves the concern relating to people spamming. Since someone could just repeatedly provide the right boolean to keep having this post the message in chat.

JstnMcBrd:
Not sure spamming is a big concern, especially considering you could do exactly the same thing with Ze Kaiser's tags system and it's never been a problem there. Besides, admins can deal with spammers individually, either by "smiting" or banning them.

AverageHelper:
Will run tests after #84 merges

Plyb:
I'm not super concerned about people spamming with this, I think it's fine to make it non-ephemeral or optionally emphemeral

src/commands/emoji.ts
```diff
@@ -0,0 +1,34 @@
...
+	);
+
+export const emoji: GlobalCommand = {
+	info: builder,
+	requiresGuild: false,
+	async execute({ reply, interaction }): Promise<void> {
+		const emojiName = interaction.options.getString(EmojiName);
+		if (emojiName === null) {
+			throw new Error('You must provide an emoji to search for');
+		}
```
AverageHelper, line 20: Since we mark the option as required in the builder, we can pass `true` as the second arg and we don't have to handle the non-existent case where the value isn't provided!

```suggestion
		const emojiName = interaction.options.getString(EmojiName, true);
```

Plyb (dismissed):

src/commands/emoji.test.ts
```diff
@@ -0,0 +1,65 @@
...
+		} as unknown as TextInputCommandContext;
+
+		mockEmojiName.mockReturnValue('existing emoji');
+		mockEmojiURL.mockReturnValue(testEmojiURL);
+	});
+
+	test('Throws an error when no input is provided', async () => {
+		mockEmojiName.mockReturnValue(null);
+		await expect(emoji.execute(context)).rejects.toThrow('You must provide an emoji to search for');
+	});
```
AverageHelper, line 43: This test may be redundant, as TypeScript assumes (ostensibly safely) that this getter will never return `null` if we pass `true` for the second argument. Discord should enforce that all required args are provided.

AverageHelper (changes requested):

TStansel:
Just pushed up a commit that adds an option in the slash command for the user to provide a boolean for when to respond ephemerally. I put the default as true and made this optional. Happy to change it if needed though

src/commands/emoji.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export const emoji: GlobalCommand = {
+	info: builder,
+	requiresGuild: false,
+	async execute({ reply, interaction }): Promise<void> {
+		const emojiName = interaction.options.getString(EmojiName, true);
+		const shouldRespondEphemeral =
+			interaction.options.getBoolean(ShouldRespondEphemeral) !== null
+				? interaction.options.getBoolean(ShouldRespondEphemeral, true)
+				: true; // This slash command defaults to sending an ephemeral message
```
AverageHelper, line 29: This is a perfect use case for [nullish coalescing](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing):

```suggestion
		const shouldRespondEphemeral = interaction.options.getBoolean(ShouldRespondEphemeral) ?? true;
```

AverageHelper (changes requested):

src/commands/emoji.ts
```diff
@@ -0,0 +1,40 @@
...
+	info: builder,
+	requiresGuild: false,
+	async execute({ reply, interaction }): Promise<void> {
+		const emojiName = interaction.options.getString(EmojiName, true);
+		const shouldRespondEphemeral = interaction.options.getBoolean(ShouldRespondEphemeral) ?? true;
+
+		const emojiCache = interaction.options.client.emojis.cache;
+		const foundEmoji = emojiCache.find(emojiElement => emojiElement.name === emojiName);
+		if (!foundEmoji) {
+			throw new Error(`Emoji ${emojiName} was not found`);
```
AverageHelper, line 31: This should probably be a `UserMessageError`, as our command executor treats other kinds of errors as a command execution failure. Since this command accepts arbitrary strings, this case is likely to happen often.

```suggestion
			throw new UserMessageError(`Emoji ${emojiName} was not found`);
```

Remember to add `import { UserMessageError } from '../helpers/UserMessageError';` somewhere reasonable in the import list, too.

AverageHelper (changes requested):

AverageHelper (approved):
Beautiful! Once #84 merges, we can run the tests and merge!

JstnMcBrd (approved):

PR #85, opened 2023-04-02
feat: Simplify our internal command options interface

AverageHelper:
I don't think we need to update changelog for this one. The only change is in DX stuff, where we're trying to get at command options. Since interaction options aren't really index-based, and we generally access the native `interaction.options` object anyway, we may as well move that north, into the command context, and get rid of the old `Array`-based options list.


src/commands/talk.test.ts
```diff
@@ -40,13 +40,13 @@ describe('Talk Slash Command', () => {
...
+		mockGetInteger.mockReturnValue(null);
 	});
 
 	test('Fails if no options are provided', async () => {
-		context = {
-			...context,
-			interaction: { options: { data: { length: 0 } } },
-		} as unknown as TextInputCommandContext;
+		mockGetInteger.mockReturnValueOnce(null);
+		mockGetString.mockReturnValueOnce(null);
```
JstnMcBrd, line 49: Given that in `talk.ts`, there is an explicit test for `if (options.data.length === 0)`, then wouldn't replacing this prevent that line from getting test coverage?

AverageHelper: Lemme check...

JstnMcBrd: I agree that this new system is better though, so if anything, that `if` statement should be removed/changed.

JstnMcBrd (approved):

Plyb (approved):

PR #86, opened 2023-04-05
Updated version of vulnerable undici dependency (pt. 2)

AverageHelper:
Based on #75, except we also fix our Jest config.

Thanks @josephgeis for the help on the dependency update, and credit goes to https://github.com/nodejs/undici/issues/1878#issuecomment-1421484960 for pointing out the solution for Jest.


SpencerHastings (approved):
🦅

Plyb (approved):

PR #87, opened 2023-04-05
Minify with rollup

AverageHelper:
A cool experiment in minifying our build. Eventually, I wanna roll up `discord.js` and `@prisma/client` and their subdeps as well, but there's some work yet to do before they'll both sit nicely in either a CJS or ESM bundle. The dream is to compile _all_ of our dependencies into a single portable JS file that can be separately deployed or distributed.

## Notable changes

- Use Rollup and ESBuild to create a minimized bundle (lets us do fancy build-time stuff later)
- Add `vitest` eslint plugin
- Use `toBe` instead of `toEqual` in unit tests
- Moved `tsconfig` settings to `tsconfig.prod.json`


AverageHelper:
Draft since #88 is higher priority

AverageHelper:
Low-key waiting on #93 and #100

JstnMcBrd:
Thanks for the `vitest` recommendation. Been using it on some of my other projects. Love the built-in TypeScript support.

I'd encourage you to create a separate PR to migrate from `jest` to `vitest` right now, even if ESM isn't ready. It's worth it.

AverageHelper:
Will do. This subproject's been on a back burner for a bit, but I have some time and motivation soon-ish to swap out the test harness.

I love Vitest so much :D

JstnMcBrd:
Apologies for pushing to your branch. I suspected merging our changes together would be messy, so I wanted to do it myself. Took a while 😅 I recommend pulling down the changes before adding more.

package-lock.json
```diff
@@ -1,7 +1,7 @@
 {
 	"name": "csbot",
 	"version": "0.12.1",
-	"lockfileVersion": 2,
+	"lockfileVersion": 3,
```
JstnMcBrd, line 4: Is there a benefit to using version 3 over version 2? I'm not familiar with the difference.

AverageHelper: Only difference is this is what NPM generated for us.

AverageHelper:
Kinda rude to force-push onto someone else's draft branch, lol

AverageHelper:
I restored some of the old tsconfig structure, but kept it simple this time.

The `"files"` and `"references"` keys in the main tsconfig.json tell VSCode which config to use in test vs prod files. The previous layout caused tests to not recognize our global types.

src/main.ts
```diff
@@ -1,5 +1,6 @@
 import 'source-map-support/register.js';
 import 'dotenv/config';
+import './helpers/parseArgs.js'; // load CLI args
```
JstnMcBrd, line 3: Is this necessary, considering `parseArgs` is already imported in `events/ready.ts`?

AverageHelper:
Closing this PR for now. I might try again one day, but now that we're Dockerized half of the point of rolling a bundle into a single file is now handled by Docker. If ever I get around to optimizing the image size or something, I may try Rollup again, or even Vite.

PR #93, opened 2023-04-12
Prefer Undici over Axios

AverageHelper:
Since [Node 18 includes the `fetch` API](https://nodejs.org/en/blog/announcements/v18-release-announce#fetch-experimental) (and that support is [no longer experimental](https://stackoverflow.com/a/75221620)), we can drop a dependency when we upgrade. Since Node's `fetch` implementation is `undici`, I think it would be prudent to use that one for now. Once we're on Node 18, we can remove `undici` (or [move it into devDependencies for type information](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/60924#issuecomment-1504635244)) for a smaller bundle.

Axios is a fine package, as far as I can tell, and very popular, even since Node 18. The `fetch` API is also popular (see [`node-fetch`](https://github.com/node-fetch/node-fetch) and the browser-native [`fetch`](https://developer.mozilla.org/en-US/docs/Web/API/fetch) API). I consider both to be industry standard REST client implementations. Since Node has added the `fetch` API natively, thus presenting a chance to drop an external dependency, I'd prefer to go that route.

## Notable Changes
- Replaced `axios` with `undici`, in preparation for Node 18.
- Used `superstruct` to create type guards at API boundaries.
- Log HTTP status messages along with non-200 status codes.

### DIFF
--- a/src/helpers/HttpStatusCode.ts
+++ b/src/helpers/HttpStatusCode.ts
@@ -1,0 +1,409 @@
+/**
+ * @see https://gist.github.com/scokmen/f813c904ef79022e84ab2409574d1b45
+ */
+
+/**
+ * Hypertext Transfer Protocol (HTTP) response status codes.
+ * @see https://en.wikipedia.org/wiki/List_of_HTTP_status_codes
+ */
+export enum HttpStatusCode {
+	/**
+	 * The server has received the request headers and the client should proceed to send the request body
+	 * (in the case of a request for which a body needs to be sent; for example, a POST request).
+	 * Sending a large request body to a server after a request has been rejected for inappropriate headers would be inefficient.
+	 * To have a server check the request's headers, a client must send Expect: 100-continue as a header in its initial request
+	 * and receive a 100 Continue status code in response before sending the body. The response 417 Expectation Failed indicates the request should not be continued.
+	 */
+	CONTINUE = 100,
+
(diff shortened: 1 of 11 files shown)


Plyb (approved):

Plyb (approved):

src/helpers/HttpStatusCode.ts
```diff
@@ -0,0 +1,409 @@
+/**
```
JstnMcBrd, line 1: This is a lot of information we may never use. Is there not a dependency that includes type codes so we don't have to define them ourselves?

AverageHelper: There are several, but none as complete as I wanted. The top two results on NPM didn't include HTTP 530, which I have observed from `findRoom` myself.

This file is largely copypasta from other projects of mine. At some point, I plan to either deploy it as an NPM package, or contribute to an already-popular package.

AverageHelper: It'd be real nice of Node packaged constants for at least the standard HTTP codes and their names, but alas, they do not.

AverageHelper: Also, I like my hover text to be descriptive, rather than needing to google a ref for HTTP status codes, lol

PR #96, opened 2023-04-15
13 starboard

### LINKED ISSUE #13: Starboard
A starboard is a channel where the bot will repost messages that have gotten a certain number of a particular reaction. The bot will post a copy of the message, along with the author, a link to the original message, a count of how many of the reaction are on the post, and what channel the message was in. When the post reaches the set threshold, it will be added to the starboard channel. From that point on, the reaction count will be updated any time someone adds or removes the specified reaction. Posts that reach the threshold then dip below it are _not_ removed from the starboard.

`/set-starboard <channel> <threshold> [emoji]` sets which channel starboard is posted in, how many reacts are required, and what emoji is being tracked (defaults to ⭐).

Multiple starboards may be set per guild. If `set-starboard` is used on a channel that already has a starboard, the old settings will be overridden. This should allow for changing the threshold without causing a "reset" in the board (i.e. all the old starboard messages should remain and continue to be updated, and new ones should not be posted if that message was already on the board). For simplicity, changing the emoji of a starboard is not supported.

If an emoji is currently being used on any starboard in the guild, the bot prevents users from reacting to their own messages with that emoji.

If an admin deletes a message in a starboard, and the source message recieves a new reaction, the bot re-posts the message.

Plyb:
Resolves #13


README.md
```diff
@@ -174,7 +174,7 @@ ADMINISTRATORS=COMMA,SEPARATED,ID,LIST
 
 ### Invite your bot to your server
 
-Go to https://discordapi.com/permissions.html#378091424832 and paste in your bot's client ID to get an invite link.
+Go to https://discordapi.com/permissions.html#379165174848 and paste in your bot's client ID to get an invite link.
```
Plyb, line 177: This is to add reaction management permissions. We'll need to re-invite Cappy after this is merged, and anyone with a a dev bot might also need to reinvite the bot.

src/events/messageReactionAdd.ts
```diff
@@ -1,43 +1,11 @@
-import * as logger from '../logger';
-import { chances, DEFAULT_CHANCE } from '../constants/reactionDuplication';
 import { onEvent } from '../helpers/onEvent';
+import { allReactionHandlers } from '../reactionHandlers/add';
+import { buildExecute } from './messageReaction';
 
 /**
  * The event handler for emoji reactions.
  */
 export const messageReactionAdd = onEvent('messageReactionAdd', {
```
Plyb, line 8: I split out reaction handling into reaction adding and reaction removing

src/main.ts
```diff
@@ -22,6 +22,7 @@ export async function _main(): Promise<void> {
 			GatewayIntentBits.DirectMessages,
 			GatewayIntentBits.GuildMessageTyping,
 			GatewayIntentBits.GuildVoiceStates,
+			GatewayIntentBits.MessageContent,
```
Plyb, line 25: This requires an update in the developer portal to allow this Intent, but without it, the bot can't read messages to make a starboard post from them.

Plyb (commented):
I made a significant architecture change by adding the concept of reaction handlers. Instead of all reaction handling being done in `messageReactionAdd.ts`, that file now simply calls the execute message on separate ReactionHandlers. So duplicating reactions was pulled out into its own handler and starboard reaction handling is also in its own handler. In addition, there is a separate messageReactionRemove.ts. Both of these can consume reaction handlers, so we can reuse the starboard reaction handling for both.

ZYancey (approved):
Looks good!

README.md
```diff
@@ -95,6 +95,10 @@ Searches for open rooms on BYU Campus. _'Now'_, _'At'_, and _'Between'_ allow yo
 
 Tracks a statistic for the issuer. Use the `track` subcommand to begin tracking, `update` to add or subtract to it, `list` to show all the stats being tracked for the issuer, `leaderboard` to show the users with the highest scores for a stat, and `untrack` to stop tracking a stat for you.
 
+### /setreactboard
+
+Creates a new reactboard or updates an existing one. A reactboard is a channel where the bot will repost messages that recieve a specified number of a specified reaction. The primary use is for a starboard where messages that recieve the right number of stars will be added, along with how many stars they recieved.
```
AverageHelper, line 100: recieved ~> received

```suggestion
Creates a new reactboard or updates an existing one. A reactboard is a channel where the bot will repost messages that recieve a specified number of a specified reaction. The primary use is for a starboard where messages that receive the right number of stars will be added, along with how many stars they received.
```

AverageHelper (changes requested):

prisma/schema.prisma
```diff
@@ -10,6 +10,27 @@ datasource db {
...
+	@@unique([guildId, channelId], name: "location")
+
+	reactboardPosts reactboardPost[]
+}
+
+model reactboardPost {
+	id Int @id @default(autoincrement())
+	reactboardId Int
+	reactboard reactboard @relation(fields: [reactboardId], references: [id])
+	originalMessageId String
```
AverageHelper, line 30: As well as an `originalMessageId`, I would like if we also include an `originalChannelId`, since Discord has no firm guarantee that a message ID is unique outside of its channel.

src/reactionHandlers/remove.test.ts
```diff
@@ -0,0 +1,12 @@
+import { _add, allReactionHandlers } from './remove';
+
+describe('allReactionHandlers', () => {
+	test('index is not empty', () => {
+		expect(allReactionHandlers.size).toBeGreaterThan(0);
+	});
+
+	test('commands can be added', () => {
+		// eslint-disable-next-line no-empty-function, @typescript-eslint/no-empty-function
+		expect(() => _add({ execute: () => {} } as unknown as ReactionHandler)).not.toThrow(TypeError);
```
AverageHelper, line 10: ESLint isn't happy with arrow functions that end in `{}`, in part because it's unclear whether the intent is a function that returns `void` (or `undefined`) or a function that returns an empty object. Returning `undefined` explicitly makes this clear to our tooling:

```suggestion
		expect(() => _add({ execute: () => undefined })).not.toThrow(TypeError);
```

src/reactionHandlers/remove.ts
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Adds a handler to the list of all reaction handlers.
+ * Only public for testing purposes. Do not use outside this file or its tests.
+ * @param handler The handler to add
+ * @private
+ */
+export function _add(handler: ReactionHandler): void {
+	if (_allHandlers.has(handler)) {
+		throw new TypeError('Failed to add handler when a command with that name was already added');
```
AverageHelper, line 21:
```suggestion
		throw new TypeError('Failed to add handler that was already registered');
```

src/reactionHandlers/add.test.ts
```diff
@@ -0,0 +1,12 @@
+import { _add, allReactionHandlers } from './add';
+
+describe('allReactionHandlers', () => {
+	test('index is not empty', () => {
+		expect(allReactionHandlers.size).toBeGreaterThan(0);
+	});
+
+	test('commands can be added', () => {
+		// eslint-disable-next-line no-empty-function, @typescript-eslint/no-empty-function
+		expect(() => _add({ execute: () => {} } as unknown as ReactionHandler)).not.toThrow(TypeError);
```
AverageHelper, line 10:
```suggestion
		expect(() => _add({ execute: () => undefined })).not.toThrow(TypeError);
```

AverageHelper (changes requested):

src/@types/ReactionHandler.d.ts
```diff
@@ -0,0 +1,12 @@
+import { MessageReaction, PartialMessageReaction, PartialUser, User } from 'discord.js';
```
AverageHelper, line 1:
```suggestion
import type { MessageReaction, PartialMessageReaction, PartialUser, User } from 'discord.js';
```

AverageHelper (changes requested):

src/reactionHandlers/updateReactboard.ts
```diff
@@ -0,0 +1,188 @@
...
+	VoiceChannel,
+} from 'discord.js';
+import { db } from '../database';
+import { appVersion } from '../constants/meta';
+
+export const updateReactboard: ReactionHandler = {
+	async execute({ reaction, user }) {
+		const fullReaction = await reaction.fetch();
+		const fullMessage = await reaction.message.fetch();
+		const fullUser = await user.fetch();
```
AverageHelper, line 22: There's a good chance that the reaction and user data aren't partial, so doing a full `fetch()` on these would be needless overhead.

```suggestion
		const fullReaction = reaction.partial ? await reaction.fetch() : reaction;
		const fullMessage = reaction.message.partial
			? await reaction.message.fetch()
			: reaction.message;
		const fullUser = user.partial ? await user.fetch() : user;
```

src/reactionHandlers/updateReactboard.ts
```diff
@@ -0,0 +1,188 @@
...
+				await fullMessage.channel.send(
+					`${user.toString()}, you can't use that react your own messages!`
+				);
+				await reaction.users.remove(fullUser);
+				return;
+			}
+
+			await updateExistingPosts(fullReaction, fullMessage);
+			await addNewPosts(fullReaction, fullMessage);
+		}
```
AverageHelper, line 51: Early return means less Christmas Tree:

```suggestion
		if (!reactboardExists) return; // abort if no reactboard

		if (fullMessage.author.bot) {
			await fullMessage.channel.send(
				`${user.toString()}, you can't use that react on bot messages!`
			);
			await reaction.users.remove(fullUser);
			return;
		}

		if (fullMessage.author.id === user.id) {
			await fullMessage.channel.send(
				`${user.toString()}, you can't use that react your own messages!`
			);
			await reaction.users.remove(fullUser);
			return;
		}

		await updateExistingPosts(fullReaction, fullMessage);
		await addNewPosts(fullReaction, fullMessage);
```

AverageHelper (changes requested):

src/commands/setReactboard.ts
```diff
@@ -0,0 +1,98 @@
...
+	return Boolean(str.match(/^\p{Extended_Pictographic}$/u));
+}
+
+function getCustomReact(client: Client<true>, str: string): GuildEmoji | undefined {
+	const reactId = str.match(/^<a?:.+?:(\d+?)>$/u)?.[1];
+	if (reactId === undefined) {
+		return undefined;
+	}
+	return client.emojis.cache.get(reactId);
+}
```
AverageHelper, line 98: Do we want to resolve emojis from every guild in which the bot exists? This might lead to issues down the road, especially if one guild gets rid of or changes an emote that another guild depends on. Also, most users won't be able to add that specific reaction anyway.

```suggestion
async function getCustomReact(guild: Guild, str: string): Promise<GuildEmoji | undefined> {
	const reactId = str.match(/^<a?:.+?:(\d+?)>$/u)?.[1];
	if (reactId === undefined) {
		return undefined;
	}
	return await guild.emojis.fetch(reactId);
}
```

AverageHelper (changes requested):

prisma/migrations/20230411151150_reactboard/migration.sql
AverageHelper, line 1: Do we have a stage in the `setup` command to run Prisma's migration CLI?

Plyb: There wasn't, but now there is!

src/commands/setReactboard.ts
```diff
@@ -0,0 +1,98 @@
...
+			.setName(channelOption)
+			.setDescription('The channel where reactboard posts will be posted')
+			.setRequired(true)
+	)
+	.addIntegerOption(option =>
+		option
+			.setName(thresholdOption)
+			.setDescription(
+				'The minimum number of reacts a message should receive before being put on the board'
+			)
```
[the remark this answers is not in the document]
AverageHelper: LGTM otherwise

AverageHelper (approved):

ZYancey (dismissed):

ZYancey (approved):
⭐️

JstnMcBrd (dismissed):

src/reactionHandlers/updateReactboard.ts
```diff
@@ -0,0 +1,193 @@
...
+			await fullMessage.channel.send(
+				`${user.toString()}, you can't use that react on bot messages!`
+			);
+			await reaction.users.remove(fullUser);
+			return;
+		}
+
+		if (fullMessage.author.id === user.id) {
+			await fullMessage.channel.send(
+				`${user.toString()}, you can't use that react your own messages!`
```
AverageHelper, line 46:
```suggestion
				`${user.toString()}, you can't use that react on your own messages!`
```

AverageHelper (changes requested):

AverageHelper (dismissed):

AverageHelper:
K I'm gonna look at this today finally

AverageHelper:
Additional considerations:
- Do we want reactboard messages to be deleted when their original message is deleted? Or are reactboards a deletion-resistant museum of sorts?
- The bot occasionally joins reacts, but it ignores `:star:` reacts. It should probably ignore all other reactboard emojis, too. We can probs add this later, since right now we only really need the starboard working.

src/reactionHandlers/remove.ts
AverageHelper, line 1: Should this file be called `index.ts` instead, following the pattern in `src/events/`?

src/events/messageReactionRemove.ts
```diff
@@ -0,0 +1,11 @@
...
+import { allReactionHandlers } from '../reactionHandlers/remove.js';
+import { buildExecute } from './messageReaction.js';
+
+/**
+ * The event handler for emoji reactions.
+ */
+export const messageReactionRemove = onEvent('messageReactionRemove', {
+	once: false,
+	execute: buildExecute(allReactionHandlers),
+});
```
AverageHelper, line 11: Feels weird for the `'messageReactionRemove'` event to execute `allReactionHandlers`, instead of just a `remove` handler..

src/commands/setReactboard.ts
```diff
@@ -0,0 +1,100 @@
...
+		let reactboardInfo: ReactboardReactInfo;
+		if (isUnicodeEmoji(react)) {
+			reactboardInfo = {
+				react,
+				isCustomReact: false,
+			};
+		} else {
+			const customReact = await getCustomReact(guild, react);
+			if (customReact === undefined) {
+				throw new UserMessageError('React option must be a valid reaction');
```
AverageHelper, line 59: For clarity:

```suggestion
				throw new UserMessageError('React option must be a valid emoji');
```

I don't think you can send a "reaction" as a command argument per-se, but _emojis_ are fair game :P

src/commands/setReactboard.ts
```diff
@@ -0,0 +1,100 @@
...
+			},
+			create: {
+				channelId: channel.id,
+				guildId: guild.id,
+				threshold,
+				...reactboardInfo,
+			},
+		});
+
+		await replyPrivately('Reactboard created!');
```
AverageHelper, line 86: Ideally this would say something different depending on whether the board was created or updated or destroyed, but since we don't differentiate that rn, more generic language might be best:

```suggestion
		await replyPrivately('Reactboard set!');
```

AverageHelper (changes requested):
Last comments and then I think it's ready:

src/reactionHandlers/remove.ts
AverageHelper: may also need to move its test file

AverageHelper (approved):
LGTM :tada:

PR #98, opened 2023-05-31
Alt text

AverageHelper:
Added a command to get the alt text from a message.

I'm not sold on the formatting, but this might be a good-enough first go for us to iterate on.

### DIFF
--- a/src/commands/contextMenu/altText.ts
+++ b/src/commands/contextMenu/altText.ts
@@ -1,0 +1,44 @@
+import type { Attachment } from 'discord.js';
+import { ApplicationCommandType, ContextMenuCommandBuilder } from 'discord.js';
+import { isNonEmptyArray } from '../../helpers/guards/isNonEmptyArray';
+
+export const altText: MessageContextMenuCommand = {
+	info: new ContextMenuCommandBuilder().setName('Get Alt Text'),
+	type: ApplicationCommandType.Message,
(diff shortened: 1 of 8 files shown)


SpencerHastings (approved):
🦀

src/commands/contextMenu/altText.ts
```diff
@@ -0,0 +1,44 @@
...
+				`None of the ${attachments.length} attachments in this message contains a text description.`
+			);
+		}
+
+		let response = '## Image Descriptions';
+		let idx = 1;
+		for (const attachment of attachments) {
+			const description = attachment.description?.trim();
+			if (description) {
+				response += `\n### Image ${idx}\n> ${description}`;
```
Plyb, line 35: are images the only kind of attachment that has descriptions?

AverageHelper: As far as I know. I've tried .exe, .pdf, and .mp4. Gif and other image formats get a Description field.

AverageHelper: We could change the verbiage to "Item X", in case Discord expands the scope later.

Plyb: I think something like that, or like "attachment x" would be good, but if it's working for now, I don't think it's enough to block the ticket going through

Plyb (approved):

PR #100, opened 2023-08-12
Use Node 18

AverageHelper:
Node 18 is stable now, and there are a number of features I'd like to take advantage of, such as:
- Built-in args parser (would make `yargs` obsolete)
- Built-in `fetch` API (would make `axios` and `undici` obsolete)

Also, [Node 16 hits EOL soon](https://github.com/nodejs/Release/blob/7b8e0ac1b97b4ce4057e64ae024995052408926a/README.md#release-schedule).

We've ditched `yargs` in this PR. We've also ditched `undici` as a runtime dependency, but we retain it for its types, as [`@types/node` doesn't yet understand global `fetch`](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/60924). Feedback is encouraged.

Those of us not developing in Docker will need to update our dev environments to Node at least 18.17.0. [`nvm`](https://nvm.sh) is a useful tool for this.


AverageHelper:
If #93 merges before this, then I'll go ahead and drop `undici` during the rebase.

JstnMcBrd (approved):

PR #111, opened 2024-01-18
Use new eslint flat config

JstnMcBrd:
I tried to make the new setup as close as possible to the old one to avoid unnecessary changes.

### Added
- ESLint configuration that uses the [new flat config](https://eslint.org/docs/latest/use/configure/configuration-files-new)
- Prisma client generation to Validate workflow

### Changed
- ESLint configuration to use new `@eslint/js` and `@stylistic` packages instead of deprecated rules ([read more here](https://eslint.style/guide/why))
- Applied new rules to all files
- Slimmed down `tsconfig.json` by removing options that are already enabled by default
- Target to ESNext, as recommended by vite


src/commandContext/replyPrivately.ts
```diff
@@ -9,6 +9,7 @@ import { replyWithPrivateMessage } from '../helpers/actions/messages/replyToMess
 export function replyPrivatelyFactory(
 	interaction: RepliableInteraction
 ): CommandContext['replyPrivately'] {
+	// eslint-disable-next-line @typescript-eslint/no-inferrable-types
```
AverageHelper, line 12: Maybe we should disable this rule outright. I quite like to be extra explicit with function argument types usually :3

AverageHelper: I'd still approve this PR if we don't, tho

AverageHelper (changes requested):
I'm about halfway through this review. Good stuff so far!

src/commandContext/replyPrivately.ts
JstnMcBrd: Good call

AverageHelper (approved):
Sorry this took so long. I wanted to cross-reference the new config and plugins against the old ones to make sure we weren't loosening things overmuch. I didn't get to do that in much detail, but the new config looks strict enough for our purposes for now. We can patch any holes that crop up later, as they appear.

The new flat config looks fun! I'm excited to play around with it

PR #115, opened 2024-02-06
Migrate to ESM

JstnMcBrd:
I have some experience migrating projects from CommonJS to ESM, so I decided to do it for this project.

I know Average was doing that in #87 , but that PR also involves adding `rollup` to our build, so I figure it'd be cleaner to separate those changes and make them incrementally.

### Added
- `verbatimModuleSyntax` setting for TypeScript

### Changed
- Converted `eslint.config.js` to ESM
- Ran `npx sort-package-json` to sort our `package.json` file automatically
- Updated typescript so we can use `verbatimModuleSyntax`
- Sorted all imports
- Changed `moduleResolution` to `NodeNext` and added `.js` extensions to all local imports
- All `logger.js` imports to use scoped imports instead of `* as logger`

### Fixed
- [`superstruct` types not being compatible with `NodeNext` resolution](https://github.com/ianstormtaylor/superstruct/issues/1200) by adding a type-shim file until it gets fixed
- `ts-node` not working with ESM by replacing it with `tsx`

### Removed
- `lodash` dependencies and replaced them with built-in methods
- All `// Internal dependencies` and `// External dependencies` comments (I don't know why I liked those in the first place)


AverageHelper:
It may be valuable too to set the `import/extensions` rule in ESLint, and set the `"ignorePackages"` option, to enforce file extensions in local imports.

As-is, unless TypeScript has changed its behavior in this area, I don't think our tooling checks for those statically like it should, rather waiting for malformed imports to fail at runtime.

src/@types/superstruct.d.ts
```diff
@@ -0,0 +1,256 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
+// This is a type-shim because superstruct's types do not work with NodeNext module resolution
```
AverageHelper, line 3: This is news to me! Superstruct is usually very good in ESM...

What exactly is the compiler error? Do we need `NodeNext` specifically, or would `ES2022` do the trick?

src/events/interactionCreate.ts
```diff
@@ -266,16 +263,16 @@ async function handleAutocompleteInteraction(interaction: AutocompleteInteractio
  */
 export async function sendErrorMessage(
 	interaction: CommandInteraction | ButtonInteraction,
-	error: unknown
+	error_: unknown
 ): Promise<void> {
-	const errorMessage = toString(error);
+	const errorMessage = error_ instanceof Error ? error_.message : String(error_);
```
AverageHelper, line 268: `String()` would turn the object into `'[object Object]'`, whereas `JSON.stringify` would traverse and print the object's properties:

```suggestion
	const errorMessage = error_ instanceof Error ? error_.message : JSON.stringify(error_);
```

AverageHelper (changes requested):

src/@types/superstruct.d.ts
JstnMcBrd: The compiler error is that `superstruct` has no exported properties. TypeScript cannot find the type files.

https://github.com/ianstormtaylor/superstruct/issues/1200

`NodeNext` is the recommended setting for modern node projects. If we set `module` to `ES2022`, then `moduleResolution` will default to `Node10`, which is designed for CommonJS-only importing. At least that's what I gleaned from the TypeScript docs.

From my experience, using `ES*` for `module` is for when you expect to run your code outside node, like in a browser.

AverageHelper: TIL! Guess I haven't ever read that part of the doc closely enough 😅

Def seems like a bug in Superstruct. I'm fine with leaving the shim in place until that's fixed

---

For my own future reference, the Summary section under [`es2015`, `es2020`, `es2022`, `esnext`](https://www.typescriptlang.org/docs/handbook/modules/reference.html#es2015-es2020-es2022-esnext) specifically says "Do not use for Node.js." Guess I should re-think my tsconfigs in all my other projects too, lol

AverageHelper (approved):

PR #118, opened 2024-02-12
Use new dectalk-tts package

JstnMcBrd:
The old `dectalk` package was a little sketchy - it created a child process to run an unprotected executable on the local machine, and it required special OS dependencies on Linux. And it didn't work on Mac.

This new package simply calls a web API, so the dectalk executable does not need to be run locally anymore. This should be much safer.

### DIFF
--- a/src/commands/talk.ts
+++ b/src/commands/talk.ts
@@ -17,18 +17,30 @@ import { fileSync } from 'tmp';
 
 import { info } from '../logger.js';
 
-const SPEAKERS = ['PAUL', 'BETTY', 'HARRY', 'FRANK', 'DENNIS', 'KIT', 'URSULA', 'RITA', 'WENDY'];
+export enum Speaker {
+	Paul = 'PAUL',
+	Betty = 'BETTY',
+	Harry = 'HARRY',
+	Frank = 'FRANK',
+	Dennis = 'DENNIS',
+	Kit = 'KIT',
+	Ursula = 'URSULA',
+	Rita = 'RITA',
+	Wendy = 'WENDY',
+}
 
 const builder = new SlashCommandBuilder()
 	.setName('talk')
-	.setDescription('Uses Dectalk to speak the given message')
+	.setDescription('Uses Dectalk to speak the given message (via https://tts.cyzon.us/)')
 	.addStringOption(option =>
 		option.setRequired(true).setName('message').setDescription('The message to speak')
 	)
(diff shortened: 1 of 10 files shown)


src/commands/talk.ts
```diff
@@ -11,28 +11,25 @@ import {
...
 import { writeFileSync, createReadStream } from 'node:fs';
 import { fileSync } from 'tmp';
 
 import { info } from '../logger.js';
 
+const SPEAKERS = ['PAUL', 'BETTY', 'HARRY', 'FRANK', 'DENNIS', 'KIT', 'URSULA', 'RITA', 'WENDY'];
+
 const builder = new SlashCommandBuilder()
 	.setName('talk')
 	.setDescription('Uses Dectalk to speak the given message')
```
AverageHelper, line 24: Might be a good idea to make clear in the command description too that we send the data elsewhere:

```suggestion
	.setDescription('Uses Dectalk to speak the given message. (Input will be processed at https://tts.cyzon.us)')
```

or maybe 

```suggestion
	.setDescription('Uses Dectalk via https://tts.cyzon.us to speak the given message')
```

src/commands/talk.ts
```diff
@@ -42,42 +39,40 @@ export const talk: GlobalCommand = {
...
-		throw new Error('Message cannot be empty');
-	}
-
 	// Generating the audio can take a while, so make sure Discord doesn't think we died
 	await prepareForLongRunningTasks(false);
 
+	// If the speaker is defined, prepend the message with the speaker's name
+	if (speaker) {
+		message = `[:name ${speaker}] ` + message;
+	}
```
AverageHelper, line 71: We should assert that the given `speaker` is one of the values in `SPEAKERS`.

AverageHelper (changes requested):

src/commands/talk.ts
```diff
@@ -42,42 +51,45 @@ export const talk: GlobalCommand = {
 	async execute(context) {
 		const options = context.options;
 		const message = options.getString('message', true);
-		const speakerNum = options.getInteger('speaker') ?? undefined;
+		const speaker = (options.getString('speaker', false) as Speaker | null) ?? undefined;
```
AverageHelper, line 54: Feels weird to not check the value of `speaker` at runtime, and assume Discord will do the filtering for us, but I'll allow it ig

AverageHelper (approved):

PR #127, opened 2024-07-10
docs: Use more specific server invite

AverageHelper:
Link the bot-specific server invite, instead of the one for the main server.


JstnMcBrd (approved):
Are doc changes worth adding to our changelog?

AverageHelper:
Probs not, they'll never be user-facing behavior

PR #129, opened 2024-07-13
Upgrade to vitest v2

JstnMcBrd:
### Added

- Generic types to all `vi.fn()` methods

### Changed

- `vitest` dependencies to v2
- Pre-existing `vi.fn()` and `Mock<>` generic types to match the new v2 types

### Fixed

- Type issues from giving generic types to `vi.fn()` methods
- Invalid node exit code CI failure by updating `prisma`
  - See this [stackoverflow question](https://stackoverflow.com/questions/76138323/the-code-argument-must-be-of-type-number-error-in-nestjs-prisma)
  - https://github.com/prisma/prisma/pull/18899

PR #131, opened 2024-07-17
Cleanup tests

JstnMcBrd:
Follow up to #129 

---

### Added

- Generics for all `Mock<>` types

### Removed

- Unnecessary mock implementations and return values, like `undefined`
- `.toBeUndefined()` assertions on methods that return `void` or `Promise<void>`
  - Usually we're just testing that they don't throw an error, so `expect()` is unnecessary
  - For rejected promises, the error is more readable when not wrapped in an `expect()`


AverageHelper (approved):
Good stuff! Thanks for cleaning this up ❤️

PR #132, opened 2024-08-13
Replace `dotenv` with Node built-in `.env` file support

JstnMcBrd:
https://nodejs.org/en/blog/release/v20.6.0#built-in-env-file-support


AverageHelper (approved):
Dang, TIL this is a thing!

PR #136, opened 2024-08-13
Refactor linting

JstnMcBrd:
### Changed

- `vscode-eslint` flag config support to be stable
- `eslint.config.js` to be type-checked
- TypeScript rules to apply to all files
- Stylistic rules to use config builder utility
- Flat configs to be inserted on the root config
- Legacy configs to be wrapped individually
- Lint rules to be organized
- All eslint-related dependencies to be updated
- Enforced new lint rules from `prettier` and `unicorn`

### Removed

- Some unused or unnecessary lint rules


ZYancey (approved):

AverageHelper (approved):
Good stuff!

PR #161, opened 2024-08-24
fix(ci): Try SSH push instead

AverageHelper:
Seems user tokens don't work for mirroring like this. Trying a Deploy Key instead.

Also checkout all history for the push, lol

### DIFF
--- a/.github/workflows/mirror.yml
+++ b/.github/workflows/mirror.yml
@@ -13,6 +13,8 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
+        with:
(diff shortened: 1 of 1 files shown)


JstnMcBrd:
Is the only way to test this to merge it into main?

AverageHelper:
That would be easiest. Afaik all the action does is do a git push using the keys we give it. I've never used Deploy Keys in a git forge before, just my account-level SSH keys. I'm just kindof hoping this works lol

JstnMcBrd (approved):

AverageHelper:
One more thing: #162

PR #166, opened 2024-08-24
feat: Publish docker

AverageHelper:
We might as well publish our own Docker image. Would make deployment much easier if a host could simply write `ghcr.io/byu-cs-discord/csbot` in a `compose.yml` file, without having to clone and build directly.

New Docker deploy flow:
- Every push to our main branch goes to an image tag called `main`
- Every new SemVer repo tag goes to an image tag called `x.x.x`, `x.x`, `x`, and `latest`
- `latest` is equivalent to the newest SemVer tag, while `main` is equivalent to the main branch
- Latest SemVer image can be pulled from `ghcr.io/byu-cs-discord/csbot:latest`
- We might additionally publish to [our Codeberg mirror](https://codeberg.org/BYU-CS-Discord/CSBot) the same way, once I know this proof-of-concept works and I figure out how to matrix the secret key

Startup time is a bit slow during command deployment, but we can speed that up later with a smarter deployment solution (one that _edits_ existing commands rather than strict redeploying, which I think would fix the issue with command permissions resetting each time).

Note also that we run database migrations on startup, too.


.github/workflows/deploy-docker.yml
```diff
@@ -0,0 +1,57 @@
...
+      contents: read
+      packages: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Log in to container registry
+        # docker/login-action@v3.3.0
+        uses: docker/login-action@9780b0c442fbb1117ed29e0efdff1e18412f7567
```
JstnMcBrd, line 30: Why the SHA instead of the normal version?

AverageHelper: Paranoia about the publish flow

AverageHelper: The odds are incredibly slim that a malicious actor would re-tag the docker actions to a malicious commit, but pinning to the hash avoids that entirely

.github/workflows/deploy-docker.yml
```diff
@@ -0,0 +1,58 @@
...
+            type=semver,pattern={{version}}
+            type=semver,pattern={{major}}.{{minor}}
+            type=semver,pattern={{major}}
+
+      - name: Build and push Docker image
+        # docker/build-push-action@v6.7.0
+        uses: docker/build-push-action@5cd11c3a4ced054e52742c5fd54dca954e0edd85
+        with:
+          context: .
+          file: ./Dockerfile
```
JstnMcBrd, line 55: Important to note that our current Dockerfile was designed to be a standardized dev environment (for everyone using Dev Containers in vscode), not a production deployment. It should probably be rewritten with the new purpose in mind.

AverageHelper: Kk, I'll get started straightening that out

AverageHelper: Ok that should to the trick: Separate Dockerfile for dev and prod, and I actually tested the prod one this time lol.

It auto-creates the database for us when using Docker! How nice

AverageHelper:
Ok I think it's done now lol. Pending new reviews

JstnMcBrd (approved):

package.json
```diff
@@ -22,7 +22,7 @@
 	"main": "./dist/main.js",
```
JstnMcBrd, line 22: Also, might be good to rebase on main before merging.

Dockerfile
```diff
@@ -1,12 +1,28 @@
-FROM node:20-slim
-RUN apt update
-RUN apt upgrade --yes
+FROM node:alpine as builder
```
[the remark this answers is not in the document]
AverageHelper: Habit I think. We should probably pull `node:20-slim` here

Dockerfile
```diff
@@ -1,12 +1,28 @@
...
+COPY package*.json ./
+COPY prisma prisma/
+COPY scripts/launch_in_docker.sh .
+
+RUN npm ci --omit=dev
+
+# Path internal to container; use `volumes` config to specify real system path of `/db/`
+ENV DATABASE_URL="file:/db/db.sqlite"
+
+RUN npm run db:migrate
```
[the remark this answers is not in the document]
AverageHelper: Good catch! I guess we should run migrations in `launch_in_docker.sh` too, instead of here (which I think happens on _build_). I'll push a fix for this and the image in a few mins

Dockerfile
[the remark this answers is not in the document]
AverageHelper: Mmm, now I'm getting issues with missing libssl lol. Gonna try `20.17-bullseye` instead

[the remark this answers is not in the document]
AverageHelper: ... or I guess I could try installing the one package, so we aren't shipping a massive image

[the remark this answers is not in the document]
AverageHelper: Fixed! Now using `node:20-slim`, and installing new `openssl` on build

Dockerfile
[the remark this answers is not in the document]
AverageHelper: Updated, and edited the PR description with that note.

AverageHelper:
Looks like the package was marked private lol, but now that's fixed and I've confirmed that the `ghcr.io/byu-cs-discord/csbot:main` image works! Once #165 merges then we should have semver tags and `:latest` as well!!

PR #178, opened 2024-09-05
Switch to SQLite

AverageHelper:
It's nice to have all persistent data in a single file, for portability. This will be useful too when setting up Docker, since upgrades and moves will be much simpler if the only things to move or modify is a compose file and a SQLite database file.


README.md
```diff
@@ -220,27 +220,12 @@ $ npm run setup
...
-- ph_host = The IP of the server running your Postgres instance (default localhost)
-- pg_port = The Port assigned to your Postgres instance (default 5432)
-- pg_db = The Name of the database you wish to use for the bot
-
-The first time you run this project, you should run the following command to initialize the database:
+Edit this line in your `.env` file:
 
 ```
-$ npm run db:init
+DATABASE_URL=file:/path/to/your/database.db
```
JstnMcBrd, line 228: Can we add instructions on how to set up SQLite?

AverageHelper: Ohh, good point! Looks like `db:init` may still be needed. Fixed.

JstnMcBrd (approved):

PR #191, opened 2024-10-03
chore: Also tag sha hash in Docker

AverageHelper:
Added another Docker tag config to target specific versions. Not sure why v0.13.0 failed to deploy, so this new tag run might give me more info


JstnMcBrd (approved):

AverageHelper:
what the heck, still only the `push main` event triggered!? At least we have tags for git hashes now, but like _aaa_ why tags no push 😭

PR #511, opened 2026-02-26
Use `clientReady` event instead of `ready`

JstnMcBrd:
Fixes a deprecation warning

```
(node:19852) DeprecationWarning: The ready event has been renamed to clientReady to distinguish it from the gateway READY event and will only emit under that name in v15. Please use clientReady instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
```


AverageHelper (dismissed):
LGTM, but needs untangling with `main`

AverageHelper (approved):

PR #541, opened 2026-04-07
Use native node for release script & eslint config instead of tsx/jiti

JstnMcBrd:
Type-stripping is stable and unflagged in Node `>=22.18.0`.

<br>

Currently ESLint requires an experiment flag to use native Node type-stripping to load the config file, despite Node type-stripping not being experimental anymore. Hopefully this is resolved soon. 
- https://github.com/eslint/eslint/issues/19985

<br>

The [`devEngines`](https://[docs.npmjs.com/cli/v11/configuring-npm/package-json#devengines](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#devengines)) field in `package.json` enforces runtime requirements for dev environments (ex: the minimum node version necessary to run `eslint`).

PR #562, opened 2026-05-13
Use native node for running the bot

JstnMcBrd:
Follow up to #541 

Modernizes and simplifies our setup

### Changes

- Fixed the required node version
  - Type-stripping is not unflagged in all version of Node 23, so it should be excluded
- Rewrote all imports from `.js` to `.ts`
- Enforced `erasableSyntaxOnly`
  - This means not using enums - only downside
- Pointed node directly to `src/main.ts`
- Removed the build step
- Replaced it with a type-checking script
- Removed all references to `dist`
- Updated the `Dockerfile` to reflect build changes
- Excluded test files from type-checking and typed linting
  - There's always been a ton of unaddressed type errors in our test files, but the build step never caught them because it excluded test files
  - The type-checking step doesn't exclude test files, so all the type errors are visible now
  - There's way too many to fix right now, so I'm saving it for later


JstnMcBrd:
The branch protection rules need to be updated to expect `CI / Type check` instead of `CI / Build`.

package.json
AverageHelper: If we go this route, we can probably do away with `source-map-support` too, since we no longer have a "build" stage to generate those.

Dockerfile
```diff
@@ -1,33 +1,31 @@
...
 RUN npm ci
 RUN npm run export-version
-RUN npm run build --omit=dev
 
-FROM node:24-slim as runner
+FROM node:24-slim AS runner
+
+ENV NODE_ENV=production
 
-RUN apt-get update -y
```
AverageHelper, line 16: Not sure about removing this line. Shouldn't we update the package lists before trying to install packages?

AverageHelper:
I'm not sure on removing the "build" stage yet. I'm hoping to eventually set up something like Rollup so we can tree-shake and avoid including node_modules in the final bundle, and setting that up would require reinstating that build step and possibly undoing this entire PR later :sweat:

Dockerfile
JstnMcBrd: Only when you want the most recent package versions, which is unnecessary for our use case (simply installing any version of openssl). Relying on auto-updates from `node:24-slim` is sufficient.

Each line of a Dockerfile is an additional layer, so I figure it's better to cut it.

JstnMcBrd:
> I'm not sure on removing the "build" stage yet. I'm hoping to eventually set up something like Rollup so we can tree-shake and avoid including node_modules in the final bundle, and setting that up would require reinstating that build step and possibly undoing this entire PR later :sweat:

Hmm. I'm not sure what value minifying, tree-shaking, or bundling would have for a project like this. It's not a web-loaded script, web app, or a library/package where download size+speed matters, and performance gains would likely be marginal.

In comparison, the simplicity + reduced cognitive load + improved devx of simply running the script directly seems very valuable. What do you think?

AverageHelper (approved):
LGTM

PR #607, opened 2026-07-14
fix: use dockerignore to exclude copied files

AverageHelper:
The CI complained that `--exclude` isn't a thing, but this might work instead.


JstnMcBrd:
Ah, it appears --exclude is an experimental feature (and has been for years), despite being included in the official dockerfile documentation.

- https://github.com/moby/buildkit/issues/2853#issuecomment-2850659442
- https://docs.docker.com/reference/dockerfile#copy---exclude

JstnMcBrd (approved):

PR #608, opened 2026-07-15
fix(docker): update package lists to find openssl

AverageHelper:
Ok so I finally actually tested the dockerfile locally (with Podman, not Docker proper, so :crossed_fingers:) and it looks like this time it has trouble finding the `openssl` package, so I suppose we must update the package lists when we run the install.

I also added a CI step to attempt building the Dockerfile, so we can catch this next time we break it, lol

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,6 +71,25 @@ jobs:
       - name: Type check
         run: npm run type-check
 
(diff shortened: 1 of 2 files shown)


JstnMcBrd (approved):
Sigh, another case of works-on-my-machine. LGTM.

